PR #13, opened 2019-01-29
manage-groups.rst: completely overhauled

martinvonwittich:
I noticed that this part of the documentation seemed subpar during this discussion: https://community.zammad.org/t/what-does-assign-follow-ups-do/2010/2

So I went ahead and overhauled it. Someone please double check that I'm really right about the Email part :)

- Improved spelling and grammar
- Improved formatting, added proper headings
- Clarified that "Support" and "Sales" are only examples, and not fixed
- Removed the cabinet/folder analogy, IMO it was unnecessary
- Mentioned the permission system and tags in the "What are groups" section
- Clarified all options, especially:
	- Assign follow ups: clarified that "leave open" means "will be unassigned"
	- Email: removed the "That means that all tickets sent to this e-mail address will be assigned to this group", it seems completely wrong to me? AFAIK Channels -> Email -> Accounts -> Destination Group decides which email will be routed to which group. Otherwise it would also be pretty much undefined what would happen when you assign the same email address to multiple groups.
	- Note: replaced the nonsensical text
	- Active: clarified that groups cannot be deleted
- Added a Permissions heading to the section dealing with permissions
- Fixed the list in the Permissions section

### DIFF
--- a/manage-groups.rst
+++ b/manage-groups.rst
@@ -1,48 +1,68 @@
 Groups
 ******
 
-This is the group-manage area. You can edit existing and add new groups.
+This is the group management area. You can edit existing groups and add new groups.
 
-What are groups actually?
-  Groups in Zammad are similar to working groups that deal with different topics within a company.
-  For example, The tickets relevant to the sales department are available in the 'Sales' group, while the tickets for the support department are available in the 'Support' group.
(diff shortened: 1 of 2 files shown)


MrGeneration:
Thank you very much for your contribution!
I just merged it into master - the changes will be available within the next 15 minutes.

PR #14, opened 2019-02-02
manage-scheduler.rst: completely overhauled

martinvonwittich:
As mentioned in [this discussion](https://community.zammad.org/t/scheduler-add-note-to-ticket/1930), I'm pretty certain by now this page was plain wrong about the note field, so I've decided to overhaul it. Please double-check though if I'm really right about this :)

- Improved spelling and grammar.
- Improved formatting, added proper headings.
- Fixed the description of the note field - it doesn't add notes to the affected tickets, but only allows the admin to save a description of the scheduler.
- Added a warning that it's not safe to delegate scheduler permissions to normal agents.
- Explicitly labeled the documented fields, e.g. 1. now says "Name". That should help when searching for the field names and should improve accessibility.
- Fixed a typo that's actually in Zammad's user interface: effected -> affected. This still needs to be fixed in Zammad itself.

What still could be improved:
- I'm not sure how I feel about the screenshots with the example scheduler. The screenshot is not accessible, so visually impaired people cannot actually tell how the scheduler in the screenshot is configured. Maybe it should also be described in text, but then it would be duplicate. I've tried to add an alt text to the screenshots, but I cannot figure out how to add a multi-line alt text in RST.
- The note "something must be done about these tickets" in the screenshots stems from the original author's misunderstanding of the note field's purpose and should probably be replaced by a note that makes it clear that it only describes the scheduler.

### DIFF
--- a/manage-scheduler.rst
+++ b/manage-scheduler.rst
@@ -27,7 +27,7 @@ Add a new scheduler
 
 The scheduler shown in the screenshot would have the following effects:
 
-Every workday (Monday to Friday) at 9:00 a. m., all tickets which:
+Every workday (Monday to Friday) at 9:00 a.m. (*UTC*), all tickets which:
 
 - are not closed and
 - are assigned to the Sales group and 
--- a/manage-scheduler.rst
+++ b/manage-scheduler.rst
@@ -1,9 +1,14 @@
 Scheduler
 *********
 
-The sheduler performs time-based automated actions. Such an action may be a change of any information on the tickets.
+The scheduler performs time-based automated actions. You can set up your own schedulers, configure at which points in time they should run, set up conditions to determine which tickets they should affect, and then configure the actions that you want to be executed on these tickets.
 
-That's how it works:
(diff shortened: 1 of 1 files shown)


MrGeneration:
Thanks for your contribution!

I added "(*UTC*)" behind the time clause, as it wasn't clear that the scheduler is running in UTC. We didn't declare this before as well, but I felt it's important to know. :)

Changes should be available within the next 15minutes!

martinvonwittich:
> I added "(UTC)" behind the time clause, as it wasn't clear that the scheduler is running in UTC. We didn't declare this before as well, but I felt it's important to know. :)

Oh wow, I didn't even notice that. I would suggest that you add that in the reference part as well, not only in the example:

> When should the job run: choose the points in time when the scheduler should run.

to:

> When should the job run: choose the points in time when the scheduler should run. Note that the scheduler works in UTC, and not in the local time zone.

Though I kinda feel like this is a bug in the scheduler. Especially considering that the example uses 9 a.m. - that feels like a time chosen to represent when Emma starts to work, and so it should be time zone dependent, and not in UTC. Maybe we could add a select field to the scheduler where admins can choose whether it should work in UTC or in the local time zone? That would enable backwards compatibility - old schedulers would have UTC configured, and new schedulers would default to the local time zone.

MrGeneration:
This is already known and is covered by: https://github.com/zammad/zammad/issues/457

PR #16, opened 2019-05-27
channels-email.rst: improved postmaster filter desc

martinvonwittich:
- explicitly explained that postmaster filters match email headers
- explicitly explained that filters do not only affect ticket creation, but also may update existing tickets
- expanded the examples and added a spam example
- explained that some actions will not have any effect on ticket updates
- added a hint that regex matching is possible
- explained that matching is always against the full header, which is especially important when using anchored regexes

### DIFF
--- a/channels-email.rst
+++ b/channels-email.rst
@@ -32,8 +32,39 @@ While creating a new e-mail account, it is possible to use the expert settings t
 
 Filters
 -------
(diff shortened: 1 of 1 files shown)


MrGeneration:
Thank you very much for this! I did some very small changes, checked the code and merged it. You'll find it online within the next 15 minutes. :-)

PR #17, opened 2019-06-16
Revised admin manual

KaldungJr:


rkaldung:
@MrGeneration Most of it was pair reading ;-) Good to go so far.

MrGeneration:
Thanks for all the work!
I double checked and also checked that everything "compiles" correctly. :-)

PR #18, opened 2019-06-20
Add Knowledge Base article

rlue:
Prepared in tandem with [this pull request](https://github.com/zammad/zammad-user-documentation/pull/21) on the Zammad Agent docs. These articles reference each other, so it would be wise to publish them together.

Further changes may be required, pending use case screenshots from @MrGeneration for a sample "Chrispresso" user.

See a live demo [here](https://private.ryanlue.com/zammad_admin/manage-knowledge-base.html).


MrGeneration:
Done!
Merging

Thanks for the amazing work, Ryan!

PR #31, opened 2019-09-30
Fix generic-cti.rst spelling error

Jaykju:
In line 44 it should be CTI API intro.


MrGeneration:
Thank you very much for your contribution! :)

PR #36, opened 2019-10-26
Fix commands for fetchmail/procmail

MichiK:
The examples are missing `zammad run` before the `rails` command.

See zammad/zammad#2791.

### DIFF
--- a/channel-email/fetchmail.rst
+++ b/channel-email/fetchmail.rst
@@ -9,7 +9,7 @@ To get this to work you need to pipe your emails to rails.
 
  su - zammad
  cd /opt/zammad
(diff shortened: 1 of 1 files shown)


MrGeneration:
Personally I'd prefer to keep redundant information at a minimum, would it be "bad" if we'd link to the following page saying "to learn how to run Zammad commands see xyz" ?

We got this covered quite detailed here: https://docs.zammad.org/en/latest/admin-console.html

MrGeneration:
I'll be closing this pull request as it didn't move up to now.
However, your feedback isn't lost - to keep the commit history clean I added the suggestions I provided here on the following merge request. 
https://github.com/zammad/zammad-admin-documentation/pull/52

I hope you're okay with this compriomise.

Issue #2, opened 2018-01-03
Update Facebook integration documentation


thorsteneckel:
Hi @KoJie - could you please update the Zammad Admin documentation on how to integrate Facebook as a channel? We received several questions (https://github.com/zammad/zammad/issues/417 / https://community.zammad.org/t/facebook-app-fehlermeldung/155 ) and should update our documentation accordingly. You can find the new screenshots attached. Feel free to ask any questions.

[Facebook Zammad.zip](https://github.com/zammad/zammad-admin-documentation/files/1599856/Facebook.Zammad.zip)

```[tasklist]
### Tasks
- [x] Setup of new Zammad system
- [x] Check Meta documentation
- [x] Add a new app at Meta's developer page
- [x] Restructure documentation section / add limitations & prerequisites
- [x] Setup facebook channel in Zammad
- [x] Add screenshots
```

MrGeneration:
I am currently moving / organizing a new order within system admin and admin documentation, thus the borders get clearer.

"Parent"-Issue to this is: https://github.com/zammad/zammad-documentation/issues/43

PR #48, opened 2020-03-23
Update trigger documentary

MrGeneration:
This pull requests is a rework of the currently existing trigger documentation:
https://admin-docs.zammad.org/en/latest/manage/trigger.html

It divides the different points into several parts to reduce page size and improve overview.
Also I changed all jpegs to png for better quality.

@rlue could you please QR the changes?

Another hint: I know I've added something on the calendar page, but I did not rework it. It needs rework (I've noted that one down), but is -in my opinion- out of the scope of this PR. :)


MrGeneration:
Small addition: I'm not entirely happy with the over of the links in the navigation. I had to learn that rst priotizes toctrees over headings and thus puts the linked documents from the toctree on top.

If you know a super hacky way how to change that, it would make me a very happy panda.
My desk will thank you as well. :D

manage/calendars.rst
```diff
@@ -1,7 +1,7 @@
 Calendars
 *********
 
-A calendar is required to calculate escalations or reports based on business hours.
+A calendar is required to calculate escalations, reports based on business hours or for trigger execution.
```
rlue: This new version of the sentence is a little messy. The intended structure is:

> **Before:** A calendar is required to calculate (escalations OR reports)  
> **After:** A calendar is required (to calculate (escalations OR reports) OR for trigger execution)

### One solution

Clarify the structure with the right punctuation and conjunctions ("and"/"but"/"or"-type words):

> A calendar is required to calculate escalations or reports based on business hours, or for trigger execution.

In other words, "A calendar is required (to calculate (escalations `or` reports) `, or` for trigger execution)."

### A better solution

The suggestion above is not ideal because some parts of the sentence are wordy (like "reports based on business hours"), and that makes it hard to see the overall structure.

A better way would be to simplify the structure of the sentence so that it's easier to see at a glance:

```suggestion
A calendar is required to:

* automate ticket escalations,
* generate reports that only capture activity during business hours, or
* set up time-sensitive triggers.
```

(And just to make sure I understand you: are calendars required to execute _any_ triggers? or only to execute the new time-sensitive triggers introduced in v3.3?)

manage/trigger.rst
```diff
@@ -1,112 +1,65 @@
...
-Maybe you want to set a higher priority on any ticket with the word “urgent” in
-the title. Maybe you want to *avoid* sending auto-reply emails to customers
-from certain organizations. Maybe you want mark a ticket as “pending” whenever
-someone adds an internal note to a ticket.
+   /manage/trigger/conditions-and-actions
+   /manage/trigger/examples
+   /manage/trigger/other-notifications
 
-Whatever it is, you can do it with **triggers:** actions that watch tickets for
-certain changes, and then fire off whenever those changes occur.
```
rlue: Is there a reason we cut this intro down and moved what's left of it to the "Default Configuration" section? It's important, IMO.

I see two ways someone will arrive at this docs page:

1. **By referral.** Someone emails support and says, "Hi, I'd like Zammad to do X when Y happens. How can I set that up?" You respond, "With triggers, dumdum! Here, check out this link."
2. **By discovery/self-directed learning.** Someone signs up for Zammad and is trying to understand all the different parts of the system. They have no mental model for how everything works in Zammad, so none of the terms make sense to them. "What the heck is a trigger? Hm, well the docs page is probably a good place to start."

In the first case, the user _already knows what problem triggers solve,_ and just needs to know more about how they work. In the second case, the user doesn't have this context, and that's what the intro is for. It's small, but it's crucial—without it, the docs become very confusing. You're reading about _how something works_ without knowing _what it does._ When that happens to me, it quickly leads to despair.

Then again, you interact with customers and I don't. Would you say that case #2 is not realistic? Or that this intro is too long and distracting for case #1?

Curious to hear your thoughts.

rlue:
I want to make sure we don't step on each other's toes, so instead of adding my own set of proposed changes to this PR, I think we should discuss it. That way, I can understand your goals (and vice versa), and we can arrive at a final version we're both happy with.

### Your goals

Reviewing this PR, it's hard to tell exactly what your goals are. I'm not saying you should have to _justify_ every single change to me, but I thought some of the text looked better _before_ the PR, and that's probably because I don't have context about why those changes were proposed.

It's also difficult to QA a large set of changes when they're for many different, unrelated goals. Here are some of the goals I see in this PR:

* Add new points to **reflect changes in the v3.3 release**—specifically, schedule-based trigger conditions and note-based actions
* **Make it easy for users to find information about "Scope"** (now the very first section of the Trigger docs, even above "How Triggers Work" 🤔)
* **Break up one long docs page into many smaller docs pages** (but keep "Scope", "Default Configuration", and "How Triggers Work" on the first page 🤔)

There are also many changes that are not related to these goals, such as the removal of the introduction or the changing of some headings (_e.g.,_ "How Do They Work?" -> "How Triggers Work"). Not sure if these are just general style edits or they relate to another goal I haven't picked up on.

If you could find some way to communicate 1) what the goals of the PR are, and 2) how each change relates to one of these goals, that would make the QA process a lot easier for me.

One way to do this would be to use cherry-picking or rebasing to clean up this branch's commit history, so that 1 goal : 1 commit, and the commit message explains what's going on.

### My goals

* I want readers to be able to **scan just the headings to figure out where they need to go**, or at least to make sure they're on the right page.

  So, for instance, "Scope" is a really bad name for a heading, because the reader has to think about it (or just read the whole section) to understand what it means. A better heading might be "Limitations (and Other Ways to Automate)".

  For this reason, I also strongly prefer question-style headings (_e.g.,_ "How Do Triggers Work?" over "How Triggers Work"). I think questions prime the reader to think about what they're going to learn, on a subconscious level. This is a personal opinion, but it's one I hold strongly. Which of the following feels better to you?

   * **What Is a Trigger?**
   * **What a Trigger Is**

* I want to **prioritize illustration over explanation** (or "show, don't tell"). 

  For this reason, I'd propose merging the "Default configuration" section into the "Examples" section. The default trigger is a great, simple example of how triggers work, and how the reader is already using them; he just doesn't know it. I've even prepared screenshots for this edit.

* I want readers to be able to **read from start to finish** without ever feeling confused.

  For this reason, I think "Scope" should go _way_ later in the article. In fact, I'd propose the following order:

  1. Learn by Example (incl. Default Configuration)
  2. Conditions & Actions
  3. Scope (or "Limitations")
  4. System Notifications

  I also think there needs to be _some_ kind of introduction, even if it's not as long as the original.

### Conclusion

Sorry, I know that was long. You asked for my QA, and you got it! :)

manage/calendars.rst
MrGeneration: Thank you very much - you're right, that sentence really is messy.
The listing is a lot better!

manage/trigger.rst
MrGeneration: Honestly, I didn't see that I removed the introduction entirely.

My goal was to reduce duplicate text and to make it "less".

So, in my opinion, most of the introduction text moved into the "scope" section. Yes, I've changed wording a lot. I thought this would enhance irritations on these parts.

In the past weeks and months I for example saw a lot of confusion on 
* when do I use a trigger?
* when do I use a scheduler?

Users get confused by them, because especially for triggers they sometimes don't work as straight forward as you might think. This is why I've been trying to work out the scope a trigger can help you. (that's also the reason I came up with the heading - but I agree, it's miss leading. )

I agree for scope, we should change it back to "How do triggers work?" to assist readers better. I didn't think it fits so I changed it, imho a bad idea. :-D

MrGeneration:
Thank you very much for your detailed feedback!

I responded to some of your points within the first QA comments to keep the threads together. I hope that helps.

You covered my goals quite good already. I wanted to add the new functionality of Zammad 3.3 in terms of triggers (and now where I think of it I forgot the scheduler part completely ;) ). 

When I was digging into the docs I was frustrated, because in my opinion it was quite messy and not what I wanted to hold. So I decided to clean it up during its extension. By doing this I was hoping to not require 2 pull requests - the first extending and the second cleaning up.

I wasn't aware that such a clean up and extension is causing such a big mess within a PR. If that doesn't help you or takes a lot more time on your end, I'll adapt and strictly try to divide those things.

---

> I want readers to be able to scan just the headings to figure out where they need to go, or at least 
> to make sure they're on the right page.
> 
> So, for instance, "Scope" is a really bad name for a heading, because the reader has to think about 
> it (or just read the whole section) to understand what it means. A better heading might be 
> "Limitations (and Other Ways to Automate)".
> 
> For this reason, I also strongly prefer question-style headings (e.g., "How Do Triggers Work?" over 
> "How Triggers Work"). I think questions prime the reader to think about what they're going to 
> learn, on a subconscious level. This is a personal opinion, but it's one I hold strongly. Which of the 
> following feels better to you?
> 
> * What Is a Trigger?
> * What a Trigger Is

As already state, I now understand the issues with the heading and would personally prefer "what is a trigger?". I think it's wording sounds better and a lot more straight forward into the direction we want to go.

---

> I want to prioritize illustration over explanation (or "show, don't tell").
> 
> For this reason, I'd propose merging the "Default configuration" section into the "Examples" section. 
> The default trigger is a great, simple example of how triggers work, and how the reader is already 
> using them; he just doesn't know it. I've even prepared screenshots for this edit.

With not being a native english speaker and proberbly a bit example for "how to learn new things" we might differ a lot on our sight of "how a documentation should look".

When supporting people, I try never to provide the full solution, because I want the person to learn. This on one side helps me, because the customer will be able to help himself better the next time,  but also the customer - he doesn't have to wait for me replying.

I'm not sure how to put it, I hope you understand what I mean.

But: I'm absolutely against merging the trigger examples and "default configuration".
For this I might have to tell you the reason why I've been adding "default configuration" last year.

This block is living on the main page (and existing), because we had to learn that users that start fresh with Zammad and add a mail account **do not expect** Zammad to send out mails in the first place.

Especially last year we had a lot of situations within the support where we had confused users complaining "they just added their productive account for testing and suddenly Zammad replied to all 5000 mails within the mailbox".

So that part is less of an example on how it works (even though our 3 default triggers provide an idea ojn how it *can* work), but more like a warning "if you do that, things will happen".

I hope that clears it up a bit for you.

---

> I want readers to be able to read from start to finish without ever feeling confused.
> 
> For this reason, I think "Scope" should go way later in the article. In fact, I'd propose the following 
> order:
> 
> 1. Learn by Example (incl. Default Configuration)
> 1. Conditions & Actions
> 1. Scope (or "Limitations")
> 1. System Notifications

The messed up order comes from me failing to "priotize" the linked aspects (so the part of the toctree) less important than the points on the index page. So the page where you start digging into the triggers. I couldn't find a way on how to re-arrange it in a way it works like I wanted to.

Personally I would suggest the following on the "trigger landing page":

1. What is a trigger? 
   Maybe we need to adjust that part a bit more to also contain the introduction you're seeking for.
   For this, we could merge "how triggers work" into the "what is a trigger / scope" part. I divided it 
   because I feld it was off. But I could have been wrong here.
1. Default configuration (excluding examples, but with the hint where to find them)

Further more, I'd like to keep the following things on sub pages to keep the size of documentation pages small:

1. Conditions & Actions
1. Learn by example
1. system notifications (a lot better wording then my decision last year!)

---

Let me know what you think, I'll gladly adapt!
Enjoy your holiday! :)

rlue:
Happy Monday! :)

> I wasn't aware that such a clean up and extension is causing such a big mess within a PR. If that doesn't help you or takes a lot more time on your end, I'll adapt and strictly try to divide those things.

Thanks, I appreciate it. If I can see "ah, this change relates to this goal," I can keep my QA less opinionated and more objective. Doesn't have to be separate PRs, necessarily; a clean commit history works, too.

---

> I now understand the issues with the heading and would personally prefer "what is a trigger?"

👍

---

> I'm absolutely against merging the trigger examples and "default configuration".

Makes total sense now that I know the context. 100% let's keep it separate and put it front and center and separate from the Examples page. 

So what would you think of highlighting the default trigger both as a warning about default config _and_ as a mini-lesson about how they work? The proposed section below could potentially replace 1) the intro, 2) _What is a trigger?_ and 3) _Default configuration._ What do you think?

> ## Triggers
>
> Use triggers to set up all kinds of if-this-then-that automation workflows.
>
> **Hint:** For every-so-often automation workflows, try the scheduler [link] instead.
>
> ### You’re already using triggers
>
> From the moment you set up Zammad, it starts sending auto-replies to all incoming emails. Recognize this line?
>
> trigger-example01_ticket_zoom
>
> It means that the ticket is from an incoming customer email, and that the customer received this message in response:
>
> trigger-example01_received_message
>
> This auto-reply message is a **trigger**. You can disable it, modify it, or create new ones for all sorts of automation tasks on the Trigger page of the admin panel:
>
> trigger-example01_admin_panel
>
> Read on for more examples. [put Examples on following page]

---

> Especially last year we had a lot of situations within the support where we had confused users complaining "they just added their productive account for testing and suddenly Zammad replied to all 5000 mails within the mailbox".

This sounds like a bad thing. Is Zammad supposed to auto-reply to old/existing emails? I know they are new _to Zammad,_ but as a user, I would only expect it to auto-reply to emails that were _sent after_ I finished setting up Zammad. Have I understood it correctly? Martin agrees this is the correct behavior?

---

I'm mostly on board with the new proposed ordering, but

> 1. Conditions & Actions
> 2. Learn by example

would prefer to flip these two (_i.e._ to favor showing vs. telling). I think people can learn more _completely_ by reading an explanation, but _faster_ by looking at an example.

The goal should be to help people get up and running fast, and then if they want to know all the details, they can go further and read the technical definitions. Then again, this is just my opinion. Does this disagree with your intuition/style of learning?

---

Learning a lot. Thanks for the pow-wow.

manage/trigger.rst
rlue: > I agree for scope, we should change it back to "How do triggers work?" to assist readers better. I didn't think it fits so I changed it, imho a bad idea. :-D

Actually, I vote "Limitations (and other ways to automate)" as a new name for "Scope". It took me a while to figure out what you were trying to say, but I do think it's different from "How do triggers work?" (I also see that it's important now!)

---

100% hear you on educating the user. Can't save everyone! ;)

I'm definitely learning to keep thing shorter. You know the pain of getting a user to read the docs, and I don't, but I think it's safe to say that long documentation doesn't get read! Mark Twain famously said,

> I didn't have time to write a short letter, so I wrote a long one instead.

So, that's something to think about.

MrGeneration:
> I'm absolutely against merging the trigger examples and "default configuration".
> 
> Makes total sense now that I know the context. 100% let's keep it separate and put it front and 
> center and separate from the Examples page.
> 
> So what would you think of highlighting the default trigger both as a warning about default 
> config and as a mini-lesson about how they work? The proposed section below could potentially 
> replace 1) the intro, 2) What is a trigger? and 3) Default configuration. What do you think?

Oh I love this approach! Didn't think of it that way.
This losens the text block up a lot and makes it way easier to understand it with less words. 
That's a great approach!


I can adjust the current branch based on your text if you want! 💪

----

>> Especially last year we had a lot of situations within the support where we had confused users 
>> complaining "they just added their productive account for testing and suddenly Zammad 
>> replied to all 5000 mails within the mailbox".
> 
> This sounds like a bad thing. Is Zammad supposed to auto-reply to old/existing emails? I know 
> they are new to Zammad, but as a user, I would only expect it to auto-reply to emails that were 
> sent after I finished setting up Zammad. Have I understood it correctly? Martin agrees this is the 
> correct behavior?

Technically this is the correct behavior, even though last week this was causing a rather shitty situation ( https://github.com/zammad/zammad/issues/3006 ). As far as I'm aware this behavior is working as designed by us to allow the user a very quick jump into the system without needing to configure too much stuff around it. 

Also, most of the times, our use case is that most of the users (if they're testing or so) don't use productive mailboxes anyway (like 98% of the people I know of I'd say). I don't want to write too much on this topic, because I might choose words that are not what I mean to say and might insult people or worse. I don't want that, sorry. :)

Another side note regarding "I'm importing all of my productive mails into a system" is that since about a year we do warn users of this situations:
image
(This message appears if you use Zammads default settings to remove mails from the inbox which could cause issues. It only appears if you have a rather "big" inbox.)

There's also a branch from Martin that will, later some day in the future™, provide "archiving" functionality incoming. This means that we will be able to import the first bunch of mails with the exact meta information of the mail it really has been received and not sendout any triggers. This is absolute future music however. I can share the said branch with you internally if you're interested to have a look.

I hope this helps you further to understand the situation.
A possible quick solution on this topic might be to either disable this trigger by default or simply extend the above posted screen with a big warning "you have triggers enabled which will do magic".

For the seeds we'd have to ask Martin on his opinion.

----

> I'm mostly on board with the new proposed ordering, but
> 
>> - Conditions & Actions
>> - Learn by example
>> would prefer to flip these two (i.e. to favor showing vs. telling). I think people can learn more 
>> completely by reading an explanation, but faster by looking at an example.
> 
> The goal should be to help people get up and running fast, and then if they want to know all the 
> details, they can go further and read the technical definitions. Then again, this is just my opinion. 
> Does this disagree with your intuition/style of learning?

Okay to me!
While it "breaks" my path of learning I think it's totally fine if it's straighter forward!

rlue:
1. :heart:

2. I understand. Personally, I really sympathize with the customer on this one—there's unexpected behavior in the system that 1) happens automatically after setting everything up, and 2) could really confuse their customers, or even potentially damage their brand image.

   Nevertheless, you guys are on the frontlines, so I defer to your expertise.

3. We can add a note at the start to make it a little easier:

   > # Learn by Example
   >
   > To get you up and running quickly, here are some examples of the kinds of automation tasks you can set up with triggers. If they don't make sense to you, don't worry—just skip ahead to the next page [link] to learn about all the options in detail, then come back here to see them in action.

---

Teamwork makes the dream work! :muscle:

MrGeneration:
Hey,

as discussed, I just updated this PR to the current state of our discussions. 
Hope it fits better now. :)

PR #52, opened 2020-05-07
Add hint on how to run zammad commands

MrGeneration:
Based on PR https://github.com/zammad/zammad-admin-documentation/pull/36 this commit adds hints on how to call Zammad console or commands on different installation types.

### DIFF
--- a/channels/email/advanced/fetchmail.rst
+++ b/channels/email/advanced/fetchmail.rst
@@ -5,6 +5,10 @@ Maybe you want to add emails via Fetchmail or Procmail to Zammad.
 
 To get this to work you need to pipe your emails to rails.
 
+.. note:: Below commands might differ from system to system. If you want to learn on how 
+  to call Zammad commands on your system, please have a look on `our console introduction 
+  <https://docs.zammad.org/en/latest/admin/console.html>`_.
+
 **Command line**::
 
    su - zammad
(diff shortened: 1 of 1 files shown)


MrGeneration:
Doh sorry, forgot to assign this to you. 
Can you please have a look when you find the time? Thanks! :)

channels/email/advanced/fetchmail.rst
```diff
@@ -5,6 +5,10 @@ Maybe you want to add emails via Fetchmail or Procmail to Zammad.
 
 To get this to work you need to pipe your emails to rails.
 
+.. note:: Below commands might differ from system to system. If you want to learn on how 
+  to call Zammad commands on your system, please have a look on `our console introduction 
+  <https://docs.zammad.org/en/latest/admin/console.html>`_.
```
rlue, line 10: After I read this note, I didn't understand

* which commands might differ from system to system,
* what systems these commands will work on, and
* how to modify these commands for other systems.

Without this information, this note raises more questions than it answers. I know the answers might be in the link, but I can't say for sure how that information relates to the code sample below.

I took a guess at what you meant and rewrote the note as follows:

```suggestion
.. note:: If you installed Zammad through a package manager (rather than from source),
   replace ``rails r`` with ``zammad run rails r`` below.
   To learn more, see `Administration via Console
   <https://docs.zammad.org/en/latest/admin/console.html>`_.
```

Did I get it right?

(Also, remember to use three-space indentation in rst directives like this. That way, the block matches up with the first line better.)

MrGeneration:
Now you're saying it...
You're absolutely right, your guess was spot on.

You can go to customer support now. 💪:-*

PR #58, opened 2020-06-14
Introduce google channel

MrGeneration:
This commit introduces the google channel for Zammad.

It makes heavy use of already existing pages of the email channel, as they are technically identically (better: the *same*) and thus don't need double work in my opinion.

I orientated myself as much as I could on the email channel.


rlue:
Looks like this will have to wait until Monday. Thanks for your patience!

channels/email/accounts/managing-accounts.rst
```diff
@@ -58,20 +58,4 @@ Enabled / Disabled
...
-         :scale: 60%
-         :align: center
-
-         Click on the address to assign it to another account, or click ✖ to delete it.
-
-      Groups need an assigned an address to send outgoing emails.
-      If you delete a group’s assigned address,
-      agents belonging to that group won’t be able to send emails
-      until you assign it a new one.
+   .. include:: managing-accounts.delete-note.rst
```
rlue, line 61: Some of these commits use the `.. include::` directive to take email pages and reproduce them word-for-word in the Google pages. I did this to make sure that the pages would never fall out of sync with each other—for example, if someone updates the email page but forgets to update the corresponding Google page.

A (potentially serious) problem with this is that all the images on these pages are hardcoded to filepaths for the email docs. In other words, now the Google article on "Managing accounts" shows images for managing email accounts.

A workaround would be to store images in the same directories as the articles that use them. This way, we could use relative paths:

```
channels/email/accounts
├── managing-accounts.01-panel.png
├── managing-accounts.02-change-destination-group-for-email-account.png
├── managing-accounts.03-updating-email-account.png
├── managing-accounts.delete-note.rst
└── managing-accounts.rst

channels/google/accounts
├── managing-accounts.01-panel.png
├── managing-accounts.02-change-destination-group-for-email-account.png
├── managing-accounts.03-updating-email-account.png
├── managing-accounts.delete-note.rst
└── managing-accounts.rst

# channels/email/accounts/managing-accounts.rst
.. figure:: managing-accounts.01-panel.png
```

channels/google/accounts/account-setup.rst
```diff
@@ -0,0 +1,123 @@
...
+=============
+
+After you’ve :doc:`registered Zammad as an OAuth app <register-app>`
+in your Google Developer settings,
+you can begin connecting Gmail accounts to Zammad.
+
+☠️ **But first, a word of warning!** The import process does things you might not expect:
+
+.. danger:: 🚯 **Zammad will delete all emails in your inbox
+   during the import process.**
```
rlue, line 11: As you can see, I put this warning back. It's good that the option is not hidden as an "expert" setting in Google channels, but it's still VERY dangerous default behavior, so I think all users should be alerted to it.

channels/google/accounts/managing-accounts.delete-note.rst
```diff
@@ -0,0 +1,6 @@
+.. note:: 🧹 **Additional Steps Required**
+
+   Groups need an assigned an address to send outgoing emails.
+   If you delete a group’s assigned address,
+   agents belonging to that group won’t be able to send emails
+   until you assign it a new one.
```
rlue, line 6: I removed the following note from Google's Managing Accounts page:

```restructuredtext
.. note:: ⚠ **No Additional Steps Required**

Unlike the Email channel, Zammad removes your configured email addresses 
together with your account if you delete it. This is, because we can reliabely 
determine them as gmail addresses which you can't use on other servers at that moment.
```

I thought about it and decided it was okay to leave undocumented: even if it works contrary to the user's expectations, there's no way it can cause any serious problems.

channels/google/accounts/account-setup.rst
```diff
@@ -0,0 +1,123 @@
...
+   Use the *Keep Messages on Server* setting to
+   :ref:`disable this behavior <google-keep-messages-on-server>`.
+
+.. warning:: 📮 **Zammad will send an auto-reply message
+   to every email it imports.** (Including the old ones!)
+
+   Make sure to :doc:`disable this behavior </manage/trigger>`
+   prior adding an email account,
+   and to turn it back on once all your messages have been imported.
+
```
rlue, line 22: I also removed the warning about how changing your Google password will invalidate your OAuth tokens. There is already so much information in that section, and I couldn't find a good, logical place to put it, so I decided it wasn't important enough to shoehorn into there.

If you think it's really important, I'm happy to put it back.

rlue:
Wait, one more important point! My changes also move the location of two files:

* `channels/email.rst` -> `channels/email/index.rst`
* `channels/google.rst` -> `channels/google/index.rst`

The reason for this is that it enables us to use the `.. include::` directive in ways we couldn't before.

### Before

```restructuredtext
.. channels/email.rst

.. toctree::

   email/accounts
   email/filters
   email/signatures
   email/settings
```

```restructuredtext
.. channels/google.rst

.. include:: /channels/email.rst
.. ^ this is broken, because its toctree points to the email pages
```

### After

```restructuredtext
.. channels/email/index.rst

.. toctree::

   accounts
   filters
   signatures
   settings
```

```restructuredtext
.. channels/google/index.rst

.. include:: /channels/email.rst
.. ^ this is works, because now the relative paths in toctree point to Google pages!
```

### The problem

This breaks the bookmarks of users who've visited the email page in the past:

* https://admin-docs.zammad.org/en/latest/channels/email.html will become
* https://admin-docs.zammad.org/en/latest/channels/email/index.html

I know you've set up 301 redirects for this situation before. Are you willing to do it again, or would you rather do away with the `.. include::` stuff?

channels/email/accounts/managing-accounts.rst
MrGeneration: Thank you so much for your effort.
Right now I'd leave the screenshot situation as be.

If we have spare time we can adjust that later on if that's okay for you. :-)

channels/google/accounts/account-setup.rst
MrGeneration: If you think it's still important enough to outline, we should keep it. 🙌

channels/google/accounts/managing-accounts.delete-note.rst
MrGeneration: I'd prefer to keep this part inside the documentation, because it took me 30-45 minutes during documentation to realise that this is working like it's supposed to be.

I'm very sure that this will, at some point, other wise cause a bug report which then relies on me remembering that this is working as intended. That's potentially dangerous.

MrGeneration:
Thank you very much for those headsup. 
I will have a look before bringing the docs live to ensure the old pages redirect as expected!

MrGeneration:
@rlue apart from the GIF issue I've completed my review - could you please have a final look on my amazing german wording? =D

Will have a look on the gif situation soonish.

PR #56, opened 2020-06-01
Added Zabbix integration do Zammad documentation

RogerTheUnicornHive:
Hello as requested in https://github.com/zammad/zammad-admin-documentation/pull/50/commits
I'm submitting Zabbix integration to the master branch of Zammad documentation.
Warm Regards,
Alex

PR #60, opened 2020-06-19
Fix Zabbix integration PR

rlue:
This commit fixes a PR which duplicated an existing article and placed it in another folder, leading to various compilation errors. See the [discussion](https://github.com/zammad/zammad-admin-documentation/pull/56#issuecomment-646416916) on #56 for more details.


MrGeneration:
Thanks, I've been fixing it on other commits :D

PR #62, opened 2020-07-07
Enhance SAML documentation

MrGeneration:
This merge request comes with a lot of different flavors.
First of all, to initially help users better, we improved two possible use cases:

* ADFS users now can find a direct hint what certificate to use (this caused confusion earlier)
* As on current support cases and a community post it became clear that "user attribute" in some cases has to be "user property" which now is stated in the docs.

To further improve the documentation, I also adjusted formatting of the page. This means I'm using more levels in order to force more structure on the page.

(hopefully) to make Ryan initially happier and reduce his load I also

* removed all white spaces that shouldn't be there
* removed blank lines not being needed

This has no super top priority - take the time you need! :)

### DIFF
--- a/settings/security/third-party/saml.rst
+++ b/settings/security/third-party/saml.rst
@@ -1,109 +1,121 @@
 SAML
 ====
 
-It is possible to create a quick login for your helpdesk via SAML for Enterprise SSO integrations.
+Connect your SAML identity provider as a single sign-on (SSO) method.
+
+.. note:: 🤷 **What is SAML?**
+
+   SAML is an open standard for SSO authentication (among other things).
+   Sign-ins are shared across multiple **service providers**
+   and managed by a central **identity provider** (IdP).
+
+   In this case, the service provider is Zammad,
+   and the IdP is a software service that you either host or subscribe to
+   (*e.g.,* `Keycloak`_, `Redhat SSO Server`_, `ADFS`_, or `Okta`_).
+
+   This guide assumes you are already using SAML within your organization (i.e., that your IdP is fully set up).
+
+   .. _Keycloak: https://www.keycloak.org/
+   .. _Redhat SSO Server: https://access.redhat.com/products/red-hat-single-sign-on
+   .. _ADFS: https://docs.microsoft.com/en-us/windows-server/identity/active-directory-federation-services
+   .. _Okta: https://www.okta.com/
+
+Step 1: Configure Your IdP
+--------------------------
+
+Add Zammad as a client/app
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Import Zammad into your IdP using the XML configuration
+found at ``https://your.zammad.domain/auth/saml/metadata``.
+
+.. note:: 🙋 **What if my IdP doesn’t support XML import?**
+
+   You will have to configure Zammad as a new client/app manually
+   using the above XML metadata file for reference.
+   For instance, when you see this tag:
+
+   .. code-block:: xml
+
+      <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="http://your.zammad.domain/auth/saml/callback" index="0" isDefault="true"/>
+
+   Set the **Assertion Consumer Service Binding URL**
+   (sometimes also listed as **Valid Redirect URIs**)
+   to ``http://your.zammad.domain/auth/saml/callback``.
+
+Set up user attribute mapping
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Zammad requests the following user attributes (or “properties”) from the IdP:
+
+* Email address (``email``)
+* Full name (``name``)
+* Given name (``first_name``)
+* Family name (``last_name``)
+
+You may need to set up “mappers” (or “mappings”) to tell your IdP
+how user attributes in SAML correspond to those in Zammad.
+For a more detailed breakdown,
+refer to the XML metadata file referenced in the previous section.
+
+Per-IdP Instructions
+^^^^^^^^^^^^^^^^^^^^
+
+Keycloak
+   * **To add Zammad as a client,**
+     save the XML configuration to disk
+     (``https://your.zammad.domain/auth/saml/metadata``)
+     and use **Clients > Create > Import** in the Keycloak admin panel.
+
+   * **To help Zammad match its own user accounts to Keycloak users,**
+     create a user attribute (or “property”) mapper:
+
+     +--------------------------------------------------------------------------------+
+     | **Clients > https://your.zammad.domain/auth/saml/metadata > Mappers > Create** |
+     +===============================+================================================+
+     | **Name**                      | EmailAddress-Email                             |
+     +-------------------------------+------------------------------------------------+
+     | **Mapper Type**               | User Property                                  |
+     +-------------------------------+------------------------------------------------+
+     | **Property**                  | emailAddress                                   |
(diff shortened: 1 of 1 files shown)


settings/security/third-party/saml.rst
```diff
@@ -19,81 +22,78 @@ Configure SAML in Zammad
...
 
 Keycloak needs to know the redirect location in advance for security:
-Goto -> Keycloak Admin -> Clients -> ``https://zammadFQDN/auth/saml/metadata``
-
-**Valid Redirect URIs:**
+Goto -> Keycloak Admin -> Clients -> ``https://{Zammad-FQDN}/auth/saml/metadata``
 
-``https://zammadFQDN/auth/saml/callback``
+Valid Redirect URIs:
+   ``https://{Zammad-FQDN}/auth/saml/callback``
```
rlue, line 82: Why does this need to be in the docs at all? I tested this on a local installation of keycloak, and importing Zammad's XML configuration file adds the Valid Redirect URIs setting automatically.

settings/security/third-party/saml.rst
```diff
@@ -19,81 +22,78 @@ Configure SAML in Zammad
...
 
 For other IdP systems, you can configure things manually.
 Zammad is using POST Bindings for the Assertion Consumer Service (ACS)
 
+ASC POST Binding URL:
+   ``https://{Zammad-FQDN}/auth/saml/callback``
```
rlue, line 88: Same question here. (The XML configuration file at `https://zammad.domain/auth/saml/callback` already includes this URL...)

settings/security/third-party/saml.rst
```diff
@@ -19,81 +22,78 @@ Configure SAML in Zammad
...
+SAML attribute via "Mappers" as needed, or equivalent mapping paradigm in other IdP systems.
+The specific attributes to map vary widely between SAML IdP systems, this requested attribute list can help align the SAML 
+provided attributes with what Zammad expects to see.
 
-Zammad requests these attributes from the SAML IdP, these are not configurable. Each attribute should map to the correct SAML attribute via "Mappers" as needed, or equivalent mapping paradigm in other IdP systems.
-The specific attributes to map vary widely between SAML IdP systems, this requested attribute list can help align the SAML provided attributes with what Zammad expects to see.
+   .. warning:: Please note that attribute may mean property in some cases! Mappings with user properties should be 
+      what you're aiming for.
 
 .. code-block:: html
```
rlue, line 99: I've rephrased this entire section thusly:

Screenshot_20200728_232715

Is that a reasonable/accurate rewriting of this section?

settings/security/third-party/saml.rst
```diff
@@ -19,81 +22,78 @@ Configure SAML in Zammad
...
+Goto -> Keycloak Admin -> Clients -> ``https://{Zammad-FQDN}/auth/saml/metadata`` -> Mappers -> Create
 
 .. code-block:: html
 
    Name: ZammadEmail
-   Mapper Type: User Attribute
+   Mapper Type: User Property
    User Attribute: emailAddress
    SAML Attribute Name: email
    SAML Attribute NameFormat: Basic
```
rlue, line 76: I've rephrased this entire section thusly:

Screenshot_20200728_234748

Is that a reasonable/accurate rewriting of this section?

#### Notes

1. In the latest version of Keycloak, "User Attribute" appears to have been renamed to "Property":

   Screenshot_20200728_233306

2. I changed `User Attribute: emailAddress` to `Property: username` because just above this, the author talks about how a) his keycloak uses email addresses as usernames, and b) Zammad needs a valid email attribute mapping. Mapping usernames to emails seemed like the logical conclusion to me... but maybe I misunderstood something?

settings/security/third-party/saml.rst
```diff
@@ -19,81 +22,78 @@ Configure SAML in Zammad
...
-
-``urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress``
-
-
-Configured example:
-
-.. image:: /images/settings/security/third-party/saml/zammad_connect_saml_thirdparty.png
-   :alt: Zammad SAML Example
-
+   ``urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress``
```
rlue, line 55: I've removed a lot of this, because large parts of it felt:

* redundant (repeating info already found in the screenshot)
* unnecessary (very long sample signing certificate)
* not authoritative; _i.e.,_ particular to some people's setups but not others, without explaining why.

Here is my rewrite of the entire section:

Screenshot_20200728_235108

Does that look correct/good to you?

#### Notes

I think the original docs specified the wrong place to copy the Keycloak signing certificate from? Happy to discuss over a screenshare.

settings/security/third-party/saml.rst
MrGeneration: I believe this is a good thing, because we explicitly this to be working with Keycloak, but "every common idP should be fine" as well. This allows you to see the information right away without having a look into the xml response.

Actually helped be a lot during a support case with a customer.

settings/security/third-party/saml.rst
MrGeneration: See https://github.com/zammad/zammad-admin-documentation/pull/62#discussion_r476344313 :)

settings/security/third-party/saml.rst
MrGeneration: Looks good to me, I'd prefer to keep
```
   .. warning:: Please note that attribute may mean property in some cases! Mappings with user properties should be what you're aiming for.
```

Because this was added during a support case I believe.

settings/security/third-party/saml.rst
MrGeneration: Your rewrite looks good and reasonable!

It's a good thing to adapt to current naming conventions or keycloak in this regard.

Regarding 2b: I believe that this won't work. The reason for this is that Zammad expects a mail address and (so I expect) does lookup a mail address in it's database. This thus shouldn't work in my understanding.

settings/security/third-party/saml.rst
MrGeneration: Looking perfectly fine to me, thank you!


> I think the original docs specified the wrong place to copy the Keycloak signing certificate from? Happy to discuss over 
> a screenshare.

I'm sorry, I can't answer this question.
The documentation has been provided by a user that was helping us with SAML.

MrGeneration:
Sorry this went below my radar and took me time to find it. :-(

settings/security/third-party/saml.rst
rlue: This phrasing okay?

> Zammad requests the following user attributes (or “properties”) from the IdP...

settings/security/third-party/saml.rst
rlue: Totally get it. How is this phrasing?

Screenshot_20200906_235701

settings/security/third-party/saml.rst
rlue: Okay. Maybe we can ask Thorsten at tomorrow's standup? Looks like he wrote the `lib/omniauth/saml_database.rb` file.

rlue: So I spoke to Thorsten, and in the end, we decided to revert it back to the original since customers have been using the existing documentation without any (reported) issues.

rlue:
Just FYI, my QA is now complete, and these docs are ready for merging pending your approval.

MrGeneration:
Looking good to me, thank you very much.
Right now I can't merge though, because approval is missing on this PR.

martini (approved):

PR #63, opened 2020-07-14
Introducing Checkmk documentation page

MrGeneration:
Hey Ryan,

this one is a bit bigger I guess. 
I tried to orientate myself on the trigger page.

On some parts I struggled keeping that style up - I hope I still got it in a okay-ish way! Personally this MR is more important than #62 on this repo (to me).

Still, no pressure on this though!


rlue:
Can you help me understand what the "Learn By Example" page is for? I see it contains a bunch of recipes for hitting the Zammad REST API, and each one is about setting some ticket attributes. Which tickets does it set these attributes on? All existing checkmk tickets? All checkmk tickets created after making this API request?

If you can use these REST API requests to set the Group on new checkmk tickets, then how is this:

Screenshot_20200717_193501

different from this?

Screenshot_20200717_193525

MrGeneration:
I initially planned to have two pages:
* available default attributes
* learn by example

But during writing the first page I noticed, that it basically already is a "learn by example", because it contains practical code on how to reach your goal. This is why it's gotten a hybrid page and may look a bit of if you take the title in account.

> If you can use these REST API requests to set the Group on new checkmk tickets, then how is this:
> 
>  image
> 
> different from this?
> 
> image

I thought my intro text of that example was good enough on this.
Technically the default group is defined within the Checkmk integration settings.

So if you post a new incident to Zammad, it will create the incident ticket in that configure group (e.g. Users).
However, if you decide that this should go into another group (for whatever reason), you can provide the `group` or `group_id` attribute to tell Zammad that the default group is wrong and you want to create the ticket somewhere else.

Maybe the information that this only works for ticket creation on that particular incident is missing.
A hint for that is hidden here: 
https://github.com/zammad/zammad-admin-documentation/pull/63/files#diff-0abdd810830a894befeb551d938d674fR79-R80

Maybe we want to provide that information on both pages.

rlue:
Ah okay. My level of understanding was even lower than I think you expected: when you said

> I initially planned to have two pages:
>
> * available default attributes
> * learn by example

the first question in my mind was, "Available default attributes _for what?_" (For Zammad's Checkmk REST API, yes, but how and when are you supposed to use these curl commands?)

After looking at it more closely, it seems that the answer is, "In the following script":

Screenshot_20200718_095744

Is that correct? If so, I think I know how I'd like to rephrase the docs.

MrGeneration:
Sorry for confusing you so much.

> the first question in my mind was, "Available default attributes for what?" (For Zammad's Checkmk REST API, yes, but how and > when are you supposed to use these curl commands?)
> 
> After looking at it more closely, it seems that the answer is, "In the following script":

This actually is something that came with Zammad 3.4: Since that version Zammad allows you to provide further attributes of **tickets** (apart from the integration ones) to set specific states of a ticket if supported.

That's a screenshot from current stable, it might help in the context. So no, this is not about the default sample script Zammad provides, but allows you to extend that use. I think I tried to keep it too short. With this you could e.g. raise the priority of an incident ticket if it's a special customer system. (This is an information that either Checkmk or the custom script needs to have)

image

rlue:
Okay, I've given it the usual treatment. Most importantly, I reorganized the docs from:

* index
  * settings
  * logging
* learn by example
* email alerts (empty)
* API alerts

to:

* index (added screenshot, shortened to basic intro)
* setup guide (new, "how does Checkmk integration work?" + sample scripts from Zammad admin panel)
* API reference
  * learn by example
  * API alerts
* admin panel reference
  * settings
  * logging

In my opinion, this structure introduces new concepts incrementally and more completely in a way that's easier for a newbie to understand. Another tiny benefit of this structure is that the sidebar nav menu now matches the actual order of the contents of the docs.

I removed some things and condensed some others, but overall, this chapter now contains _more_ content than before. I know this goes against my usual rule of wanting to make things shorter, but I felt quite strongly that this section needed more explanation: I had to do quite a bit of independent research to figure out what was going on, and of course, the docs should contain everything you need to know on their own.

Let me know what you think!

system/integrations/checkmk/setup-guide.rst
```diff
@@ -0,0 +1,98 @@
...
+   it needs to provide some information
+   about the event that triggered the notification.
+   This information is passed in the form of
+   these ``$NOTIFY_*`` environment variables.
+
+   You can specify additional parameters to pass to the script
+   when you’re setting up your notification rule,
+   but the ones you see here are all provided by default.
+
+Scripts for Email Alerts
```
MrGeneration, line 88: This heading really bothers me.
Technically while checkmk does use a script under the hood, this will be a generic email based notification sent from checkmk. This means that there's no script involved.

I'd prefer a heading like e.g. `E-Mail notification based alerts` or something like that.
Because otherwise the question would be "how do I have to write my script for those mails??" which basically would be incorrect.

system/integrations/checkmk/api-reference.rst
```diff
@@ -0,0 +1,173 @@
...
+.. note:: 🙅 The following attributes are **not customizable**:
+
+   * title
+   * id
+   * ticket number
+   * customer
+   * created_by_id
+   * updated_by_id
+
+How do I know what values I can set?
```
MrGeneration, line 142: While I love the rewrite of the above, I noticed the part below misses information from my original commit.
This especially hits the fan in terms of id based attributes.

e.g. the possibility to use either `group` or `group_id` if you like.
Depending on the flavor you prefer.

Do you think this is too much input and "named" attributed are good enough?

system/integrations/checkmk/api-reference.rst
```diff
@@ -0,0 +1,173 @@
...
+      :alt: See possible values for certain attributes in the ticket pane.
+      :scale: 50%
+      :align: center
+
+   Refer to the dropdown menus in the ticket pane:
+
+   .. code:: bash
+
+      -F "group=Users"
+      -F "state=closed"
```
MrGeneration, line 156: Attention on state!
This is a default checkmk attribute that does **not set the ticket state**!

My original commit contained this:
```
State
   You can set a specific state id to create the ticket with another state than it usually would (new).
   
      * Reference way ``state_id=2``
```

This is one of the reasons why I decided to provide the user with two options: named and id based references. State is an exception where only setting via ID is possible.

MrGeneration:
Can you please have a look on my feedback? The rest looks good to me.
You really managed to sort things out even better than I could achieve.

system/integrations/checkmk/setup-guide.rst
rlue: TIL! Obviously I am new to Checkmk. Would you be cool with it if we changed both headings to the following?

* API Alerts
* Email Alerts

system/integrations/checkmk/api-reference.rst
rlue: This is a really tricky point. I'm glad you pointed this out, because I 100% didn't catch it the first time around. I think this deserves a note; working on rewriting it now.

system/integrations/checkmk/api-reference.rst
rlue: Ah, you're right—I forgot to mention that I removed this.

I will prepare a rewrite that covers this point, but IMO this subject really deserves its own article. Understanding how to find the `group_id` for a given group requires and understanding of DB relations and the underlying structure of Zammad's data model, which is something that I wouldn't expect the average reader to know about or be able to figure out on their own. Ideally, this article could live in the "Danger zone (for experts)" section of the zammad-documentation repo, and we could link there from this page.

In reality, I don't expect that to happen anytime soon, so I'll just make a small note of it here.

system/integrations/checkmk/setup-guide.rst
rlue: This is addressed in the latest commit. Leaving this conversation open so that you can a) resolve it if you're okay with it, or b) continue the discussion if you're not.

system/integrations/checkmk/api-reference.rst
rlue: This is addressed in the latest commit. Leaving this conversation open so that you can a) resolve it if you're okay with it, or b) continue the discussion if you're not.

system/integrations/checkmk/api-reference.rst
rlue: This is addressed in the latest commit. Leaving this conversation open so that you can a) resolve it if you're okay with it, or b) continue the discussion if you're not.

MrGeneration:
Thank you very much for your rework.
Looking good to me! :-)

PR #70, opened 2020-09-15
Provide Data Privacy Documentation

MrGeneration:
Hello Ryan,

this pull request provides the all new data privacy feature. 
I already applied the suggestions you provided within Mattermost.

Please have a look at wording, structure and germanized english grammar. :-P
This pull request may not be merged before release of Zammad 3.5.

---

If you have questions, especially about the functionality and want a look and feel, please feel welcome to let me know. I'll gladly explain and show!

Thanks beforehand!

PR #71, opened 2020-09-15
Refactor or users and roles

MrGeneration:
Hello Ryan,

this pull request is quite a big.
Might have gotten a slight cluster fuck, I tried to behave! 👀🙏

I must confess that this wasn't too easy for me, because there's a lot of important information living inside. I thus copy cated the style we introduced within Google and email channel hoping it's a good path for both fast readers and those who care for details.

**Important:** This PR depends on PR https://github.com/zammad/zammad-admin-documentation/pull/70 and already links to data privacy. Just so you don't wonder why there's compiling errors. :D

---

Please have a look at wording, structure and germanized english grammar. :-P
This pull request may not be merged before release of Zammad 3.5.

If you have questions, let me know, I'll gladly answer them. We can have a call too!

Thanks beforehand!


rlue (approved):
Yowza, this was a lot of changes. I know we're in a hurry, so I'm just going to go ahead and merge now—but please do look over it and tell me if anything needs to be changed after.

manage/roles/admin-permissions.rst
```diff
@@ -0,0 +1,167 @@
...
+``admin.package``
+   Provides access to the :doc:`package management </system/packages>` for Zammad addons.
+
+``admin.report_profile``
+   Allows users to define :doc:`/manage/report-profiles` to enhance their Reports.
+
+      .. hint:: 🤓 Report profiles are relevant for the Reporting.
+         Access to the Reporting function depends on :doc:`agent-permissions`.
+
+      ..tip:: This is a great permission for key users. 🚀
```
rlue: Sorry, I did not get what you meant by "key" users, so I deleted the two or three lines that contained notes like this.

manage/groups/access-levels.rst
```diff
@@ -0,0 +1,104 @@
...
+
+“The Intern”
+   .. figure:: /images/manage/groups/access-levels/create-only.png
+      :alt: The group access table, checked "Create"
+      :align: center
+      :width: 50%
+
+   Agents with only “create” access can do just that,
+   and nothing else—once they hit **Save**, they’ll never see that ticket again.
+   Great for taking phone calls for someone more important than you.
```
rlue, line 104: Just FYI, these examples are written with a humorous tone. Happy to change them if the team feels this is inappropriate for technical docs.

manage/roles/admin-permissions.rst
MrGeneration: No worries.
I'm not sure if there's an international wording for this:

A key user in germany is a user that has a better technical understand than the other users.
This user is not an administrator but has a greater understanding of the software in use.

German companies use key users in order to spread knowledge to other users.
This helps to reduce the number of people that you have to send to a workshop and also allows you to keep on working if you send users to a workshop.

Because if the whole company goes to a workshop, that'd mean you'd have to take a "day off" support which in some cases doesn't help. I'm fine with leaving this wording away if it doesn't fit!

manage/groups/access-levels.rst
MrGeneration: I'm fine with the tone, it's not too evil.
In case this comes up sour, we'll adapt it later. 💪

manage/roles/admin-permissions.rst
rlue: It sounds like what you're describing is a “team lead”? Conversationally/half-humorously, you may refer to someone like this as the “resident expert” on a subject (but I definitely wouldn't use that term here). But to be honest, I haven't spent enough time in large teams in the workplace to know if there is really a comparable term in English.

More broadly, it sounds like we're describing another role between Agent and Admin—sort of a Super Agent? My take is that it would be better to illustrate this concept by actually having this sample role included by default in every instance of Zammad, rather than adding a note to every relevant permission in the reference docs.

MrGeneration: Okay. Let's just skip this part then for now. It's not super important. :)

PR #74, opened 2020-09-22
Enhance information regarding notification channel

MrGeneration:
Up to now the documentation doesn't declare a notification channel being dependent on the notification sender address.
This leads to confusion on new members especially in situations where you start using SMTP which restrictive mail servers that only allow specific mail addresses per user.

This pull request adds a direct hint within the channel notification detail dialogue.
Also, earlier the wording for email notification channels on the overview page was stating this channel being relevant for "auto-replies" which is miss leading. This is why I replaced it with a more fitting wording.

### DIFF
--- a/channels/email/accounts/email-notification.rst
+++ b/channels/email/accounts/email-notification.rst
@@ -26,7 +26,8 @@ Host
 User
    Your account login/username.
 
-   .. hint:: The notification sender address can be configured within the :doc:`Settings tab <../settings>`.
+   .. hint:: The “From:” address on system notifications can be configured under
+      :doc:`Channels > Email > Settings > Notification Sender <../settings>`.
 
 Password
    Your account password.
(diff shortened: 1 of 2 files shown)


channels/email/accounts/email-notification.rst
```diff
@@ -26,6 +26,8 @@ Host
 User
    Your account login/username.
 
+   .. hint:: The notification sender address can be configured within the :doc:`Settings tab <../settings>`.
```
rlue, line 29:
```suggestion
   .. hint:: The “From:” address on system notifications can be configured under
      :doc:`Channels > Email > Settings > Notification Sender <../settings>`.
```

rlue (changes requested):
Just the one suggestion. Sorry it took so long to get back to you!

MrGeneration:
No worries, thanks for having a look.
You're right, your approach is much better!

Is this ready to go from your end?

rlue (approved):
:100:

PR #77, opened 2020-09-30
Address PR71 refactor (small changes, additions)

MrGeneration:
Hey Ryan,

here's my QA related to PR#71.
Some things felt off or were missing in my opinion.

It's mostly informational and detailed nature.
Can you please have your vision on it? The rest of PR71 was totally fine! Loved the switch of the group-permissions to actual group parts.

Thanks!


manage/groups/access-levels.rst
```diff
@@ -20,7 +20,7 @@ Within each group, the different access levels allow an agent to...
 :OVERVIEW: ...see ticket overviews (but not ticket details)
 :FULL:     ...all of the above **and be assigned tickets**
```
rlue, line 21: How about

```suggestion
:FULL:     ...all of the above **and be assigned / receive notifications for tickets**
```

instead of the note below?

manage/roles/user-preferences-permissions.rst
```diff
@@ -45,6 +45,8 @@ User Preferences Permissions
 
                                        .. note:: Agents only receive ticket notifications for
                                           :doc:`groups they have “full” access to </manage/groups/access-levels>`.
+
+                                       .. hint:: Customers don't need this permission.
```
rlue, line 49: Would prefer to combine these two notes, maybe like so?

```suggestion
                                       .. note:: Agents can only receive ticket notifications for
                                          :doc:`groups they have “full” access to </manage/groups/access-levels>`.
                                          Customers can’t receive ticket notifications at all.
```

manage/roles/user-preferences-permissions.rst
```diff
@@ -57,4 +59,4 @@ User Preferences Permissions
                                        .. warning:: 🔑 **Third-party authentication / LDAP users:**
 
                                           Be sure to **revoke this permission** for all your users.
-                                          Otherwise, user passwords may fall out of sync.
+                                          Otherwise, users will have two passwords (local & LDAP).
```
rlue, line 62: Can you explain how this works? AFAIU Zammad only has one column in the users DB table for user passwords. How would users end up with two?

manage/users/index.rst
```diff
@@ -57,7 +60,7 @@ The ones that aren’t are described below.
 .. note:: 🕵️ **Admins aren’t the only ones who can change these settings.**
 
    In most cases, agents can, too
-   (using the `new ticket dialog`_ or the `ticket pane`_).
+   (using the `new ticket dialog`_, search function or the `ticket pane`_).
 
    .. _new ticket dialog: https://user-docs.zammad.org/en/latest/basics/service-ticket/create.html
```
rlue, line 65:
```suggestion
   (using the `new ticket dialog`_, `search function`_, or the `ticket pane`_).

   .. _new ticket dialog: https://user-docs.zammad.org/en/latest/basics/service-ticket/create.html
   .. _search bar: https://user-docs.zammad.org/en/latest/basics/find-ticket/search.html
```

manage/users/index.rst
```diff
@@ -33,12 +33,15 @@ Learn more about managing users...
...
-   🪄 **poof**—new customer account!
+   ✨ **poof**—new customer account!
 
    (Your customers never need to set a password.
    Of course, they can if they want to,
    but the account will be there even if they never use it.)
 
+.. hint:: Zammad users are global to the whole instance. 
+   Restriction to specific groups is not possible.
+
```
rlue, line 44: IMO this hint belongs in the groups page, not the users page.

### Why?

Putting this hint here creates more questions than it answers, simply because it's too early:

This is the first page in the entire admin docs—and it's a good first page because everyone everywhere understands the concept of users. The concept of groups, however, is specific to Zammad; it takes some time for new users to create a mental model for this concept.

Thus, it's likely that there are going to be a _lot_ of readers on this page that don't know what a group is yet; there is no way for them to understand the point you are trying to clarify. But there will never be a reader on the group page who doesn't understand what a user is yet.

manage/roles/agent-permissions.rst
```diff
@@ -17,10 +17,24 @@ Agent Permissions
...
+:``cti.agent``:             Provides access to `📞 Caller Log <https://user-docs.zammad.org/en/latest/extras/caller-log.html>`_
+
+                            .. hint:: 🤓 Requires configuration of either integrations
+
+                               .. toctree::
+                                  :maxdepth: 1
+
+                                  Generic CTI </system/integrations/generic-cti>
+                                  placetel </system/integrations/placetel-cti>
+                                  sipgate </system/integrations/sipgate>
```
rlue, line 32: Making this a `toctree` leads to these headings appearing in the sidebar TOC:

Screenshot_20201002_230756

Was that on purpose? IMO it doesn't fit the organizational structure of the document.

rlue (changes requested):
Sorry, I have opinions about everything! Please push back if you think I'm nitpicking.

manage/roles/user-preferences-permissions.rst
MrGeneration: Yes sure.
So basically Zammad doesn't require a password on your user account if you use a third party login source like LDAP.

When using LDAP, Zammad will contact your LDAP server during your authentication. It knows if the account has been synced from LDAP and thus will handle exactly that for your user. This is also the reason why the LDAP server has to be available during the moment you're trying to authenticate with a LDAP user.

Zammad does not (and never) synchronize the password of the LDAP user. 
So if I say you're having two passwords technically I do mean:

- you can have an independent, different password on the user account that lifes in Zammads database
- The LDAP source is used for your authentication try
- if you do have two passwords, one technically could fail but the other password may allow you to login
   - for instance: Your ldap server is not available due to network issues (this counts as authentication fail in terms of Zammad) - BUT if you have a local password on that user within the Zammad databae and it's correct, you're still able to login


I hope this helps to understand the scope better.

manage/users/index.rst
MrGeneration: Good point, much better. 🙈

manage/roles/agent-permissions.rst
MrGeneration: Good point. From my end it was on purpose because I technically like to see where it leads.
However: I do get your point that the structure virusally goes into the wrong direction. I'll revert this.

manage/groups/access-levels.rst
MrGeneration: Thanks for your input, will be address with commit 229e5a8c55fcf709e0eadb232d165c1959a47a63

manage/roles/user-preferences-permissions.rst
MrGeneration: Thanks for your input, will be address with commit 229e5a8c55fcf709e0eadb232d165c1959a47a63

manage/users/index.rst
MrGeneration: will be address with commit 229e5a8c55fcf709e0eadb232d165c1959a47a63

manage/roles/agent-permissions.rst
MrGeneration: will be address with commit 229e5a8c55fcf709e0eadb232d165c1959a47a63

MrGeneration:
I'll review the rest later on.
You can resolve the stuff I replied to if it's good to go for you!

manage/users/index.rst
MrGeneration: Very good point.
I moved it to groups with commit 42f7a1bf00e99487fca3f04936e60103fc294410

manage/roles/user-preferences-permissions.rst
rlue: Ack! Totally didn't realize this was sitting here waiting for a response from me. 

Great explanation. I'd like to finesse this explanation a little more, because it's technically correct, but creates more questions than it answers:

* "will users always have two passwords, or is it just that they _can_ have two passwords?"
* "are the two passwords necessarily different? what happens when they are the same?"
* "will both passwords work?"

(Of course, your explanation answered these questions for me, but we want to keep it as short as possible.)

I think a higher-level explanation of the **intent** / security model would work better:

```suggestion
                                          When using a third-party identity server (like LDAP),
                                          the whole point is to let *it* take care of authentication
                                          so that passwords never have to live in Zammad’s database.
```

MrGeneration: Ah good point. This part gently gives nose bleeding. :D

Thanks!

MrGeneration:
From my point of view this is good to go. :-)

rlue (approved):

PR #80, opened 2020-10-30
Introduce Microsoft365 Channel to documentation

MrGeneration:
This commit provides the required documentation for Zammads new Microsoft 365 Channel.
The style of the documentation orientates on the Gmail documentation to 100% and even has 70% of it's original wording.

If there's question or you need some hands on knowledge, please let me know.


channels/email/accounts/account-setup.rst
```diff
@@ -19,11 +19,15 @@ Setting up a new email account? Here’s what all the settings do.
...
-   **Office365 / Exchange users:** Shared mailboxes are not supported.
+      Google is in the process of upgrading its security policies.
+      To stay up-to-date, do **not** add your account as an email channel—instead,
+      create a :doc:`Google channel </channels/google/accounts/account-setup>`.
+   **Microsoft 365 users:**
+      Microsoft is in the process of upgrading its security policies.
+      To stay up-to-date, do **not** add your account as an email channel—instead,
+      create a :doc:`Microsoft 365 channel </channels/microsoft365/accounts/account-setup>`.
+
+   **Microsoft 365 / Exchange users:** Shared mailboxes are not supported.
```
rlue, line 30: I really like this idea of using RST definition lists in larger, more complicated notes like this one.

For better legibility/formatting, I'd prefer something like this:

```rst
.. note::

   **Gmail / G Suite users:**
      Google is in...

   **Microsoft 365 users:**
      Microsoft is in...

   **Microsoft 365 / Exchange users:**
      Shared mailboxes are not supported.
```

but GitHub won't let me add this as a code suggestion, so I'll just push it in a new commit.

rlue:
The file `account-setup/add-microsoft365-account-to-zammad.gif` appears to have some heavy ghosting. Would it be a lot of trouble to re-record it and try for less compression?

rlue (approved):
Hey Marcel,

Here are my proposed changes. I even discovered some typos in the Google docs during this process!

Mostly my changes are related to wording, but in some cases I added a verbal description to supplement a screencast/gif. (I know I always talk about how less is more, but I thought maybe if we have blind users, it would be helpful to have written instructions for accessibility reasons.)

Let me know your thoughts!

channels/google/accounts.rst
```diff
@@ -47,9 +47,8 @@ Accounts
 Email Notification
    This option is **not available on Google channels**.
 
-   To configure the outgoing mail provider for system notifications,
-   use a :doc:`regular email notification channel </channels/email/accounts>`.
-   (Self-hosted installations only.)
+   (See the corresponding section on the
```
MrGeneration, line 50: I must confess that I'm not a big fan on removing essential information (again!).
Especially the "self hosted part only" is a crucial one.

I mean I do understand you want to keep it short, but we really have to stop removing important bits.

channels/microsoft365/accounts.rst
```diff
@@ -47,9 +47,8 @@ Accounts
 Email Notification
    This option is **not available on Microsoft365 channels**.
 
-   To configure the outgoing mail provider for system notifications,
-   use a :doc:`regular email notification channel </channels/email/accounts>`.
-   (Self-hosted installations only.)
+   (See the corresponding section on the
+   :doc:`Email Accounts </channels/email/accounts>` page to learn more.)
```
MrGeneration, line 51: Same as the google channel bit.

MrGeneration:
> The file `account-setup/add-microsoft365-account-to-zammad.gif` appears to have some heavy ghosting. Would it be a lot of trouble to re-record it and try for less compression?

Yes no worries.
I had a feeling and saved all recording projects before turning them into gifs.

However, this one made me struggle quite a bit which is why I couldn't maintain the small file size.

I had to leave a lot of image steps before freezing them for a second (so that it's not jumping too quickly) in order the tool to allow removing the ghosting on its own.

Some grey parts of Zammad are not dark enough which why some button outlines are lost.
I hope you're happier with the result now. :-)

channels/google/accounts.rst
rlue: I hear you. When I came (back) to this section, I just thought there was a lot of potential for confusion. We're basically saying:

> ### Setting that's not in the control panel
>
> This setting doesn't exist. But actually, it does, just only on regular email channels. So if you want to do this thing that maybe you weren't even thinking about doing, go there instead. But actually, you can only do that if you're self-hosted; on cloud-hosted instances, this setting _really_ doesn't exist. Like, at all.

and all that back-and-forth was just too much for me.

Importantly, I'm making an assumption about the questions in the reader's mind when they're on this page. My assumption is that they're thinking, "Hm, how do I set up a Microsoft 365 account?" and not "Hm, how do I configure my Microsoft 365 account to be the address for system notification emails?"

But as always, you know the customer (i.e. support caller) better than I do. I'll prepare an edit now and push that to this PR for your approval.

channels/microsoft365/accounts.rst
rlue: Addressed in 0a03e47. Let me know what you think!

MrGeneration:
That's so much better!
Thank you very much. 🎉

PR #82, opened 2020-11-15
Introduce email archive import mode

MrGeneration:
Hello Ryan,

sorry for being so late with that.
This Pull Request handles the upcoming "archive import" function of the email channel.

You can only set that up if:
* you have mails older than 2 weeks in your mailbox
* you choose imap to be your fetch protocoll
* you choose "keep messages on server" to be "false"

Only mails prior the channel creation date will be imported as archive (state = closed, no notifications at all (neither trigger nor agent). All mails newer than that creation date will be imported as regular messages.

I hope this helps.
Currently this is not available for google and microsoft 365 channels - however, I did use the microsoft365 channel branch just in case this becomes available soonish.

I don't have hope in that regard though.

If you need anything else, please let me know!

**This PR is blocked by `microsoft365-channel` PR.


channels/email/accounts/account-setup.rst
```diff
@@ -170,6 +178,35 @@ Keep messages on server
...
+   .. note:: 
+
+      🤓 The archive option only affects tickets before the channel creation date.
+      All newer mails will be imported as ``regular``.
+
+.. figure:: /images/channels/email/account-setup-archive-import.png
+   :alt: Optional archive import mode for email channel.
+   :align: center
+   :width: 60%
+
```
rlue, line 209: I have a proposed rewrite, but the content of this PR looks totally redundant with the warning in the screenshot of the email account setup dialog. In my opinion, this PR should remove L14-L20 and leave out this new section entirely.

If you'd be okay leaving this section out, I'd prefer to go that way—though I understand that sometimes you need to be able to send users a link to a page in the documentation to help answer their questions.

---

```suggestion
Import as
   During import, Zammad normally 📮 **sends an auto-reply
   to every single email in your inbox**
   (including messages you’ve read already, even from months or years ago!)

   Use this option to control this behavior.

   .. figure:: /images/channels/email/account-setup-archive-import.png
      :alt: “Import as” option in Email account setup dialog
      :align: center
      :width: 40%

      How should old emails be imported?

   archive
      Import old messages (from 2+ weeks ago) as “closed” tickets.
      Do not send auto-reply emails, notify agents, or fire triggers for these tickets.
      These tickets will not appear in any of the `default overviews <https://user-docs.zammad.org/en/latest/basics/find-ticket/browse.html>`_;
      use the search bar to locate them.

   regular
      Import every message in the inbox as a “new” ticket.
      Send auto-reply emails, notify agents, and fire triggers
      as if each message were received today.

```

rlue (changes requested):

channels/email/accounts/account-setup.rst
MrGeneration: You're right, it partly is redundant.
However:

On the new archive import part I'm missing this (important) note:
```
.. note:: 
   This option only appears if you have emails older than two weeks in your mailbox, 
   configured Zammad not to keep messages on the server and fetch mails via IMAP.
```

Because this option is not always available. I feel this is an important part to ensure we don't get bug reports on something the documentation should have stated before.

Also, for Line 14-20 I see your point - however, it only is redundant in case you turn on archive import.
It's not if you keep the messages on your server or still decide to regularly import your mails.

It's also a good hint for users that do not expect this default behavior - simply to be aware of it!

---

Not sure, maybe Line 14-20 needs a rewrite, but I wouldn't want to remove it. :x

What's your thought?

rlue: > On the new archive import part I'm missing this (important) note:

I knew I was forgetting something! Thanks for pointing that one out.

I just pushed a new set of proposed changes. Now back to you!

MrGeneration: Wonderful, much better - thank you! 🤲

rlue (approved):

PR #93, opened 2021-02-19
Rework Macros & Introduce relative time functionality for Macros & Triggers

MrGeneration:
This pull request reworks the current Macro page to fit our standards (it has a similar feeling like our Trigger section) and also cleans it up. It's dependent on this PR https://github.com/zammad/zammad-user-documentation/pull/43 which takes over the user part.

After reworking Macros, I've added information about the relative time functionality.
It's a fairly small addition but should do the trick.

I'd like to suggest to take the changes as whole which may save time on your end. :-)

---- 

Note:
Relative times is still missing on the scheduler documentation.
The reason for that is that the scheduler section needs a rework which is out of the scope of this PR.
This currently is a time issue on my end which is why I cut that off. We'll add that later.


rlue:
Hey Marcel! Here are my proposed changes. As usual, I got a little carried away. Important notes:

* I removed the "Limitations" page. I felt strongly that that information on it did not need its own page, and that it was confusing to say, in four different places, "See Limitations for more details."
* I also removed all the lines that said "In some situations, macros may be the wrong choice," because I believe it creates more questions than answers.
* I also removed the reference to relative time in **Macros > How do they work?** because I thought the reference was too short to be useful. In other words, either explain what relative time is and how it works, or don't say anything and let the user figure it out on their own.

I hope I didn't forget anything. Let me know what you think of the changes!

rlue:
Oh wait, just remembered, also removed a tip/hint that said "This document describes how to **configure** macros. For more on how to **use** them, refer to the Zammad user docs."

This sentence was lifted from the KB docs. I think it was more important for the KB docs because using the KB also involves a lot of _editing,_ which feels similar to admin system configuration, and which is not the case at all for using macros. In my opinion, if we include that notice on the macros page, we should include it on virtually every page in the admin docs.

rlue (approved):
Approved with my latest QA commits, but (as always) open for further critique and discussion.

MrGeneration:
Thanks for your effort!
Your changes look good for me.

I'm also happy you put the limitation hints (both) to places it does fit.

PR #94, opened 2021-03-09
Extend knowledgebase editor permission with hint

MrGeneration:
As it might be confusing why a editor automatically receives reader permission as well (without ticking those in), we've added a remark on this behavior.

@Martin - could you please check if my wording is good enough for this case? :)
Do you miss anything?


martini (approved):
Works fine for me! Can get merged.

PR #97, opened 2021-03-22
Provide GitHub and GitLab Integration documentation

MrGeneration:
This pull request provides both GitHub and GitLab integration pages. 
I've decided to put both versions into one pull request, because they're quite similar technically (and in wording) which probably is easier to overview.

Let me know if you need something.


rlue (approved):
Approved with revisions from latest commit.

rlue:
Note that I removed the following warnings from the GitHub article:

```rst
.. warning:: Below permissions to not work with private repositories. 
   GitHub does not provide granular read only for those which is why 
   you have to use ``repo`` permission instead. We don't recommend that - 
   it provides a lot of additional access.
```

Per the official GitHub docs, the `public_repo` scope is not read-only, either—it offers read-write permissions, just not for private repositories.

Before I found those docs, based solely on the warning above, I got the wrong impression of how the `repo` and `public_repo` scopes worked. Do you think it's still important to warn people against using a read-write OAuth scope?

system/integrations/github.rst
```diff
@@ -0,0 +1,57 @@
...
+
+   .. figure:: /images/system/integrations/github/configure-and-active-github-integration.gif
+      :alt: Screencast showing how to configure Zammad’s GitHub integration
+      :align: center
+      :width: 90%
+
+   .. hint:: Leave the default API endpoint (``https://api.github.com/graphql``) as-is
+      unless you’re using GitHub Enterprise Server.
+
+Once completed, a new tab for GitHub issues will appear in the ticket settings. 🎉
```
rlue, line 44: I'm honestly not sure if I interpreted this part correctly. It might even help to add a screenshot here!

Do you mean to say that in the "New Ticket" dialog, there's now an option to make the ticket a GitHub issue?

MrGeneration:
> Per the official GitHub docs, the `public_repo` scope is not read-only, either—it offers read-write permissions, just not for private repositories.
> 
> Before I found those docs, based solely on the warning above, I got the wrong impression of how the `repo` and `public_repo` scopes worked. Do you think it's still important to warn people against using a read-write OAuth scope?

Thanks for the hint, may have gotten something wrong.
However, this warning is a crucial one containing *important additional* information in case of using private github repos.
I believe that this makes this piece of information important. 

This is why I'd like to suggest the following:
```
.. warning:: **Below permissions to not work with private repositories.**

   If you want to use private repos too, please use ``repo`` permission instead. 
   We don't recommend that - it provides a lot of additional access.
```

rlue:
I'm so sorry, I thought I included a link to the docs in my last comment. Here they are:

https://docs.github.com/en/developers/apps/scopes-for-oauth-apps

Experimenting with this a bit further, I discovered something interesting: for public repos, _you don't need the `public_repo` scope at all!_ Per the docs above:

> `(no scope)` Grants read-only access to public information (including user profile info, repository info, and gists)

This is sufficient to link public issues on a ticket in Zammad. (Just tried it myself.) As such, I will be amending my last commit to reflect these changes, and including a modified version of the warning you included in your last comment.

### A request

Could I ask you to re-record the screencast to show API token generation with no scopes at all? :pray:

MrGeneration:
Thanks for your time. I replaced the screencast to reflect your findings.
Another plus for that situation si that I could save 300kb over the old version. :D

PR #100, opened 2021-05-21
Create tags.rst

christophdb:
the search is "tags:xxx" and not "tag:xxx"


MrGeneration (approved):
Thanks for your contribution. 
Even through this page is generally outdated and needs rework, I'm still accepting the fix to reduce confusion. 💪

PR #118, opened 2021-10-13
Add Core Workflows Ajax-Fallback

MrGeneration:
This adds the Ajax-Fallback mode for Core Workflows to the documentation helping administrators to understand the possible benefits and scopes of the option.


settings/system/frontend.rst
```diff
@@ -1,8 +1,39 @@
...
+      and other. **Do not** enable this option unless we're asking you to.
+
+Client storage
    Use client storage to cache data to enhance performance of application.
+
+   Default: ``no`` (inactive)
+
+.. _core-workflow-ajax-mode:
+
+Core Workflow Ajax Modus
```
techerella, line 22: Since the feature is called Core Workflows (plural), should we add an "s" here?

settings/system/frontend.rst
```diff
@@ -1,8 +1,39 @@
...
+   This setting allows administrators to enforce :doc:`/system/core-workflows`
+   to use Ajax-Calls instead of web sockets. You'll usually only need this if
+   you experience serious issues as noted below.
+
+      .. hint:: **🤓 Possible (technical) reasons**
+
+         In some cases your network structure (e.g. firewalls, proxies)
```
techerella, line 29: In some cases,

settings/system/frontend.rst
```diff
@@ -1,8 +1,39 @@
...
+         may disconnect long web socket connections. This leads to
+         select fields staying empty (e.g. owner selection *after* selecting
+         your group) or fields not shown / hidden (e.g. when switching to
+         or from pending states, the pending till field not showing / hiding).
```
techerella, line 33: I would add colons: the "pending until" field not showing

MrGeneration: image

The field is actually called "pending till", but I'll add colons to help understanding that better.

PR #124, opened 2021-12-06
Update microsoft.rst

mialikescoffee:
If I understand it correctly URL must be "accountname.zammad.com/..." and not "zammad.domain.com/..."

### DIFF
--- a/settings/security/third-party/microsoft.rst
+++ b/settings/security/third-party/microsoft.rst
@@ -57,7 +57,7 @@ Supported account types:
 Redirect URI (optional):
    Select web and provide your callback url. 
    The callback url looks like this: 
-   ``https://zammad.domain.tld/auth/microsoft_office365/callback``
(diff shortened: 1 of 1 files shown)


MrGeneration:
Thank you for your contribution, however:
The `zammad.domain.tld` is just a sample domain.
It's not fixed to hosted environments but also applies to self hosted installations which is why -in my opinion- forcing it into a `zammad.com` flavor feels wrong to me. We've used `domain.tld` on several points.

What do you think?

mialikescoffee:
For self hosted environments it totally makes sense. I was just a little confused while configuring O365 authentication or SAML over Azure.  I guess most Zammad installations are self hosted anyway? Then it should be fine. :)

MrGeneration:
Ah sorry for the confusion.
In case you did not yet know: You can also hook custom CNAMEs to the hosted environment.

In case this is relevant for you, contact our support and you'll be good to go! :)

PR #128, opened 2022-01-19
Fix Keycloak attribute to map

BLeQuerrec:
Keycloak attribute to map for SAML is not `emailAddress` but only `email` (at least on Keycloak 15.0+).


ralf401:
Hi @BLeQuerrec 
thanks a lot for your contribution and sorry for the late response. Your are right, it has to be `email`.

Due to a recent change in the saml.rst file, there are conflicts now with your PR. I will close your PR, merge the changes and give you credits there.

PR #136, opened 2022-03-13
Rework group settings & add shared drafts setting

MrGeneration:
This PR adds the shared draft group setting and also reworks the settings page a little bit to make it's overview and feel better.

Note:
If everything looks good for you *without* complaints, you can directly hit the `Merge` button.
The PR is landing in a pre-release branch and does not affect life documentation pages.


manage/groups/settings.rst
```diff
@@ -3,40 +3,104 @@ Group Settings
...
+      This is the default value
+   no
+      The owner assignment of the ticket will be removed.
 
-- *Active*: choose whether the group is enabled or not. Groups cannot be
-  deleted, they can only be set to inactive.
+Email
+   Ticket groups decide on which email address Zammad will use outbound
+   for replying on a ticket. *Destination groups* of your email based channels
+   do not have any effect on this setting and may differ.
```
YetAnotherGerrit, line 49: I like the former wording for the Email field much better. I would skip the comment about destination groups, I think the number of people that may be confused by it is higher than the number of people that are helped with it at this point. Instead I would add a comment to the documentation of destination groups, that the sender is configured in the group settings.

YetAnotherGerrit (changes requested):

manage/groups/settings.rst
MrGeneration: Fun fact: There's already such comments in the email section and it's in general overseen.
This is why there's an enhancement issue for this on this repo.

Do you still think having it twice is a bad idea?

YetAnotherGerrit: Haven't checked the documentation of the Email-section to be honest. Great to see, that there is a comment about the senders address already.

But yes, still prefer the old wording here.

MrGeneration: So you want only the following?


```
Email
   Ticket groups decide on which email address Zammad will use outbound
   for replying on a ticket.

      .. note:: **You don't have an email address configured yet?**

         Please configure an email based channel before here

            * :doc:`Channels → Email </channels/email/index>`
            * :doc:`Channels → Microsoft 365 </channels/microsoft365/index>`
            * :doc:`Channels → Google </channels/google/index>`

         and come back afterwards.
 ```

YetAnotherGerrit: Suggestion:
Select which sender's email address Zammad will use outbound for replying on a ticket in this group.

Instead of:
Ticket groups decide on which email address Zammad will use outbound
   for replying on a ticket. *Destination groups* of your email based channels
   do not have any effect on this setting and may differ.

PR #169, opened 2022-07-19
Closes #88 - Make it clear that removing users alters reports as well

### LINKED ISSUE #88: add. note: DSGVO/ deleting tickets <> Reporting
zammad-admin-documentation/system/data-privacy.rst

It is often asked how the reporting behaves when tickets are deleted. Therefore, it should be added either in the body text or in the FAQ's that deleted tickets are also deleted for reporting. 
As a recommendation, so that these tickets still show up in the statistics, automation can help to set the customer to a "dummy customer".

MrGeneration:


system/data-privacy.rst
```diff
@@ -129,9 +130,13 @@ Frequently Asked Questions
...
-😵 **I deleted an user and can still see a message they sent!**
+😵 I deleted an user and can still see a message they sent!
    Tickets can only belong to a single customer,
    but may contain messages (“articles”) from many people.
    If you deleted a user but you’re still seeing articles they sent,
    don’t worry—those articles are for a ticket that belongs to someone else,
    and no longer contain any reference to the sender’s identity.
+
+🚮 I removed a customer, my reporting now is off!
+   When removing users and their ticket, all references are removed.
```
YetAnotherGerrit, line 141: "now my reporting is off" instead of "my reporting now is off".

"tickets" instead of "ticket" in row 141.

YetAnotherGerrit (changes requested):

YetAnotherGerrit (approved):

PR #188, opened 2022-11-20
Add public links functionality

MrGeneration:


mgruner (approved):
@MrGeneration please see my commits for small changes that I'd propose. Thanks!

PR #189, opened 2022-11-21
Clean up system frontend settings; Add week number setting

MrGeneration:


mgruner (approved):
Small wording suggestions provided in my commit. Thanks!

PR #191, opened 2022-11-21
Maintenance: Update screenshots of triggers

MrGeneration:
We're doing so only where the screenshot would visually miss the note field to provide full scope on screenshots.

The note field is not described as these sections provide a fairly visual approach.


MrGeneration:
I'm merging this without QA as there was no wording changes / additions.

PR #196, opened 2022-11-21
Introduce ticket templates management

MrGeneration:


mgruner (changes requested):
Small grammar/style suggestions in my commit.

adding-new-template.png has a typo too, which is even underlined by your browser: `service degration` should probably be `service degradation` (two times, in text and also in title). Please correct that and then feel free to merge!

PR #201, opened 2022-12-20
Update and improve readme

MrGeneration:
With switching branches, the readme needed an update. Also user and admin docs both have different information on the same portions which also leads in inconsistencies. This commit should solve these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Information about this markup language can be found at:
 
 This documentation provides versions:
 
-- ``pre-release`` can contain develop, not yet releases functions and changes
+- ``pre-release`` can contain develop, not yet released functions and changes
 - ``main`` is the ``latest`` (and stable) version of the repository
 - ``stable-x.x`` is the old back port for an earlier version
   - These branches do not receive further updates and serve as historic help
--- a/README.md
+++ b/README.md
@@ -2,36 +2,47 @@
 
 # Zammad admin documentation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,36 +2,47 @@
...
 - http://www.sphinx-doc.org/en/stable/rest.html
 - http://docs.readthedocs.io/en/latest/_themes/sphinx_rtd_theme/demo_docs/source/demo.html
 
-Thanks! ❤️ ❤️ ❤️
+### Versioning
+
+This documentation provides versions:
 
-  Zammad Team
+- ``pre-release`` can contain develop, not yet releases functions and changes
```
mgruner, line 35: not yet `released`

mgruner (approved):
Looks great, one minor typo.

README.md
MrGeneration: good call. Will fix that in both versions.

README.md
```diff
@@ -2,36 +2,47 @@
 
 # Zammad admin documentation
```
mgruner, line 3: Hm, we might want to capitalize this too.
But OTOH, if you don't use capitalization style consistently yet in the docs, we can keep it like it is for now.

mgruner: "Title case" that is. https://capitalizemytitle.com/

MrGeneration: No it's a good point and addressed already. :)

PR #234, opened 2023-04-28
Fixes #231 - Add more webhook examples.

### LINKED ISSUE #231: Add more webhook examples
* Mattermost Notifications
* Rocket Chat Notifications

dvuckovic:
Changelog:

* Add Mattermost Notifications guide.
* Add Rocket Chat Notifications guide.
* Add links to new guides to the Webhook Examples page.


dominikklein (approved):
@dvuckovic Looks good for me :-)

Only one thoughts in general about our new examples.

Maybe we can add one page for the trigger creation for all of these notification webhooks. Here we could add trigger examples for all events which exist in the old slack integration (create, update, pending reached, escalation, escalation warning). Then we could link this in all single webhook examples.

dvuckovic:
@dominikklein Sounds like a great idea!

But, I would recommend we create a follow-up issue and move the discussion over there. This already stands on its own, and we should apply it as soon as possible.

Could you perhaps do this? And describe proposed triggers (which attributes are needed for what behavior, etc)?

Thanks!

MrGeneration:
Up to now the documentation was supposed to give you hints and ideas on how to do stuff.
It's task was not to wipe the butt of people.

This sounds like this is going to be a huge duplicate chunk of stuff and would be a possible bad idea.
Just my 2 cents.

dominikklein:
> @dominikklein Sounds like a great idea!
> 
> But, I would recommend we create a follow-up issue and move the discussion over there. This already stands on its own, and we should apply it as soon as possible.
> 
> Could you perhaps do this? And describe proposed triggers (which attributes are needed for what behavior, etc)?
> 
> Thanks!

Yes I will create a follow-up.

dominikklein:
> Up to now the documentation was supposed to give you hints and ideas on how to do stuff.
> It's task was not to wipe the butt of people.
> 
> This sounds like this is going to be a huge duplicate chunk of stuff and would be a possible bad idea.
> Just my 2 cents.

But my idea was to simplify the current approach to have one page which all example pre-defined webhooks pages can use.

dominikklein:
@dvuckovic created https://github.com/zammad/zammad-admin-documentation/issues/237

PR #238, opened 2023-05-03
Fixes #237 - Pre-Defined Webhook Examples - Generic trigger creation page.

### LINKED ISSUE #237: Pre-Defined Webhook Examples - Generic trigger creation page
Add one page for the trigger creation for all of these notification webhooks. Here we could add trigger examples for all events which exist in the old Slack integration (create, update, pending reached, escalation, escalation warning). Then we could link this in all single webhook examples.

dvuckovic:


manage/webhook/examples/generic-notifications-trigger.rst
```diff
@@ -0,0 +1,80 @@
...
+   Select ticket **Action** attribute, leave the operator on *is* and set the
+   value to *created*.
+
+Ticket Updated ``Action activator only``
+   Select ticket **Action** attribute, leave the operator on *is* and set the
+   value to *updated*.
+
+Ticket Merged ``Action activator only``
+   Select ticket **Action** attribute, leave the operator on *is* and set the
+   value to *merged into* or *received merge*.
```
dominikklein, line 54: I think this is not needed here.

dominikklein (approved):
@dvuckovic looks good to me, I would only remove this one mention piece because I think it's not needed in this context.

PR #279, opened 2023-08-07
Replaced screenshots for triggers

ralf401:


images/manage/organizations/general-view-of-a-busy-organization.png
dvuckovic, line 1: The new screenshot seems a tad bit too large: let's recreate it in 1280x720 max.

images/manage/trigger/move-tickets-of-specific-customer-to-specific-group.png
dvuckovic, line 1: Same here: the dialog seems a bit too large.

images/manage/trigger/trigger-based-auto-replies-on-follow-ups.png
dvuckovic, line 1: Size again.

images/manage/trigger/trigger-based-priotize-change.png
dvuckovic, line 1: Size + typo in the filename.

dvuckovic (changes requested):

images/manage/organizations/general-view-of-a-busy-organization.png
ralf401: Oh it looks like my already smaller screenshot was overwritten somehow :-O

I just wanted to show the same information as before, but I think here it is no problem to limit it to 1280x720. Then the list of open / closed tickets will only show the title.

I re-upload it.

images/manage/trigger/move-tickets-of-specific-customer-to-specific-group.png
MrGeneration: Just my two cents:
There are several places in the documentations where you'll find screenshots that are drastically larger than the readme's limit at least for one side of the image. For the exact reason Ralf mentioned. When it was necessary information are shown, I usually used higher resolutions to overcome this or started to cut away uneeded UI parts like the tab bar (especially for triggers)

dvuckovic: I totally agree that in some cases the height of the screenshot has to be longer, however I would still try to create it against the width of 1280px. If we did our job correctly, all important elements should be contained in this width, and then the length only can be longer so everything of note is contained within.

As @MrGeneration mentioned, it's perfectly fine to screenshot just the relevant node (in this case a dialog), including the "chrome" of the app around it does not bear any purpose.

ralf401: I reduced it already, just showing the modal window and width is <1280px. So I will leave it as uploaded. I will double check the other ones, why they are different in width (not 1175px) and re-upload if necessary.

images/manage/trigger/trigger-based-auto-replies-on-follow-ups.png
ralf401: Recreated screenshot. You are right, the modal was created without the limitation on 1280 in width.

dvuckovic (approved):

PR #283, opened 2023-08-11
Fixes #240: Redacted Gitlab API doc

### LINKED ISSUE #240: Security - GitLab third-party part is outdated
The documentation of GitLab third-party documentation needs to be improved, because of the following issue: https://github.com/zammad/zammad/issues/4602

(https://admin-docs.zammad.org/en/latest/settings/security/third-party/gitlab.html)

ralf401:


settings/security/third-party/gitlab.rst
```diff
@@ -16,15 +16,16 @@ where zammad_host has to be replaced with your Zammad FQDN
    :alt: Register OAuth app on gitlab instance
 
 At the moment we need the "api" scope. This is caused due a bug
-in Gitlab: https://gitlab.com/gitlab-org/gitlab-ce/issues/33022
+in Gitlab.
```
dominikklein, line 19: I think we should not touch this and I will create a follow-up fix, with this no API-Scope in general is no longer needed.

locale/admin-docs.pot
```diff
@@ -12017,22 +12017,22 @@ msgid "Enter the \"APP ID\" and the \"APP SECRET\" from the Gitlab OAUTH Applica
 msgstr ""
 
 #: ../settings/security/third-party/gitlab.rst:27
-msgid "Please ensure to use ``https://{git_host}/api/v4/`` for site."
+msgid "Please ensure to use ``https://{git_host}/`` without any ``/api/v4`` suffix for site."
```
dominikklein, line 12020: I would not say something about the suffix, because for "new" installations it could be confusing, this sentence would only be relevant for older instances.

dominikklein (approved):
Only 2 small thoughts.

PR #290, opened 2023-08-24
Maintenance: fixed broken links and badges

ralf401:


mgruner (approved):
As discussed, please consider having badges for both `latest` and `pre-release`.

PR #292, opened 2023-08-25
Fixes #263 - Added note about current situation 

### LINKED ISSUE #263: Twitter-integration: update on working status
To my knowledge the Twitter-integration is not working and the future of the API uncertain. Maybe we should add a note to the part of documentation and keep that updated?

ralf401:


ralf401:
Points to pre-release => wrong => closed

PR #304, opened 2023-09-01
Fixes #301 - Added SSL/STARTTLS verification / removed ldap(s):// suffixes

### LINKED ISSUE #301: Change LDAP configuration after SSL/StartTLS changes inside the wizard
https://github.com/zammad/zammad/issues/4746

- Inside the host, the protocol needs no longer to be inserted
- Added SSL/StartTLS which is always visible
- SSL-Verify field which is visible when selecting SSL or StartTLS

ralf401:


fliebe92 (approved):

ralf401:
Feature for 6.2 => reverting commit after coordination with @dvuckovic and @mgruner until new doc-branch is created

PR #313, opened 2023-09-06
Fixes #310 - Added note for migrating old regex filters at updating Zammad

### LINKED ISSUE #310: Migration of operators/values from old regex to new at updating Zammad

ralf401:


misc/object-conditions/basics.rst
```diff
@@ -392,6 +392,14 @@ Available operators for matching:
    Triggers, Time Accounting selector, Postmaster Filters, Automatic ticket
    assignment and Core Workflow.
 
+.. hint::
+   **What about my "old" style regex filters?**
```
dvuckovic, line 396: Perhaps we should switch `regex` to something like

```
``regex:...``
```

So it's clear to the readers which old style are you refering to.

misc/object-conditions/basics.rst
```diff
@@ -392,6 +392,14 @@ Available operators for matching:
...
+
+   If you update your Zammad from 6.0 or prior and you have already conditions
+   with *contains* or *contains not* including a regex filter, Zammad tries to
+   migrate them to the new *matches regex* and *does not match
+   regex* operators.
```
dvuckovic, line 401: Again, perhaps a small `(i.e. ``regex:^(foo|bar)$``)` as an example.

dvuckovic (changes requested):

dvuckovic (approved):

PR #320, opened 2023-09-07
Draft: Fixes #316 - Added editable Base-DN field

### LINKED ISSUE #316: Possibility to use LDAP integration without allowed anonymous bind
https://github.com/zammad/zammad/issues/4471

When anonymous bind is not allowed on the LDAP side this will be detected and the "Base DN"-Field in the second wizard step will be a plain text field instead of a prefilled select field (like it is with allowed anonymous bind).

ralf401:


ralf401:
Hi @dominikklein, can you have a look if it is ok that way? 

Just a note: I had to push it in this branch because there were already changes at the LDAP section and it is not yet merged. It is just about the lines 60-65 in manage.rst (changes from today's commits).

fliebe92 (approved):

PR #337, opened 2023-09-18
Fixes #336 - Added rails commands for setting a channel to trusted

### LINKED ISSUE #336: Add console commands for trusted channels
As this in general can be a usefull thing, we should also document it.

See commit
https://github.com/zammad/zammad/commit/1e8616151c308fd9ad0dbd72b1c6ee6385054a9b#diff-836a9832d3b8bf7760d717ff5b434180R29

ralf401:


ralf401:
@mgruner can you have a look? Thanks!!

channels/email/email-headers.rst
```diff
@@ -158,3 +159,46 @@ email articles.
...
+
+.. code-block:: ruby
+
+   >> options = channel[:options]
+
+Add the ``"trusted"=>true`` flag for the inbound part of the channel:
+
+.. code-block:: ruby
+
+   >> options[:inbound][:trusted] = true
```
mgruner, line 201: This looks good, but I believe that the settings will not be saved. Try it by reopening the Rails console and checking again.
`channel.save!` should do the trick.

mgruner (approved):
Looks good, but saving is still missing and needs to be corrected.

channels/email/email-headers.rst
ralf401: Thank you! Worked :-)

PR #365, opened 2023-10-05
Maintenance: Add tip to verify S3 setup.

tschaefer:


ralf401 (approved):
Looks good! 

I just noticed one thing: the complete `.. tip` block should be indented because it belongs still to the S3 method.

PR #368, opened 2023-10-10
Fixes #364 - Added hint for workaround empty fields

### LINKED ISSUE #364: Using empty variables in signatures
*use case* 
* have one or several user attributes as text fields added via object manager (like title, department, etc.)
* use one or more of these attributes in a signature via variable
* leave the field (in user edit menu) empty and observe how "-" is substituted in the signature
* This works as designed

Nevertheless, at https://admin-docs.zammad.org/en/latest/system/variables.html we state in the hint box: 

> You have an empty field which you referenced and it appears as -? That’s currently working as designed - you might want to ensure that these fields always have a value (in text fields \`` \``  is a value!).
 
The sentence in parentheses does not make sense to me. I have tried multiple variants of this to have an empty value in the customer edit dialogue for the individual attribute (e.g.: title), so that the variable is not substituted by "-", but is left empty instead.  

What *does* work, is to use the HTML paragraph tag to achieve the desired result, i.e. not showing anything, if the user attribute is empty. 

The following value in the user attribute makes Zammad "skip" the whole user attribute in the signature when the corresponding variable is substituted: 

`<p> </p>`

We should update this accordingly in our documentation in the hint box mentioned above.

ralf401:


ralf401:
Hi @cautopartes,

thanks for your issue. Can you have a look if it is ok now?

Thank you!

PR #369, opened 2023-10-10
Fixes #361 - Edited wording for objects and attributes

### LINKED ISSUE #361: Reword the whole chapter on objects so it refers to attributes appropriately
* **Object**
  A business object in Zammad (i.e. Ticket, User, Group...)

* **Object Attribute**
  A custom field for an **object** in Zammad (i.e. Text, Boolean, Integer...)

Currently, the whole [Objects chapter](https://admin-docs.zammad.org/en/latest/system/objects.html) refers to both entities as _objects_, which is very confusing.

Consider separating the two terms, and being more specific when talking about **object attribute types** and configuration of **object attributes**. Basically, the only change that can be made to an **object** is which type of **object attributes** are defined. The actual entity that most of the chapter is trying to explain are **object attributes**, not **objects**.

However, it's perfectly fine to refer them as _attributes_ in a short form, so we can avoid too long sintagmi.

ralf401:


system/objects.rst
```diff
@@ -1,23 +1,23 @@
-Objects
-*******
+Object Attributes
```
dvuckovic, line 1: I don't think we should not change the title of the section, since it's aligned with the UI at the moment. While it's fine to introduce object attributes later in the textx, the section title + link in the sidebar should stay on `Objects`. What do you think?

system/objects.rst
```diff
@@ -1,23 +1,23 @@
...
-Here's an overview of the objects. On the upper right you can add new
-Attributes. By default, there will be no custom fields - standard objects
+Here's an overview of object attributes. On the upper right you can add new
+attributes. By default, there will be no custom fields - standard attributes
 will be grayed out, you can't delete or change those.
-Custom objects (will be displayed in black font and have a trash bin on
-the right site to delete unnecessary objects.
-By clicking on "custom objects" you can edit them to suit your needs.
+You can edit custom attributes (they are displayed in black font and have a
+trash bin on the right side) just by clicking on a it.
```
dvuckovic, line 20: `by clicking on a it.` => `by clicking on them.`

system/objects.rst
```diff
@@ -36,39 +36,39 @@ By clicking on "custom objects" you can edit them to suit your needs.
...
 If you're ready to go, just click on "Update database" to apply the changes
 to Zammad. If you made a mistake or just want to discard your changes, click
 "Discard changes".
 
 .. warning::
 
-   After applying the object changes with "Update Database" a restart of Zammad
+   After applying the changes with "Update Database", a restart of Zammad
    is **mandatory**. If you don't perform it, you may experience unexpected
    behavior or even errors. You may want to do those kind of configurations
```
dvuckovic, line 73: `those kind of configurations` => `this kind of configuration`

system/objects/types.rst
```diff
@@ -224,8 +226,8 @@ actual URL of the field - e.g. ``https://zammad.com``.
 How does this work...?!
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-As an example, let's say you have an object called ``amazingobject`` - you want
-to open a google search directly with the input from that field.
+As an example, let's say you have an attribute called ``amazingobject`` - you
```
dvuckovic, line 229: Should we also adjust the name of the example attribute? It's only logical, since it will only be saving a reference to another object anyway. Something along the same lines like `amazing_attribute` or whatever would do.

dvuckovic (approved):

system/objects.rst
ralf401: Yes I agree. However, we should consider renaming it in the Zammad UI in the future.

system/objects/types.rst
ralf401: It makes sense, yes! I changed it to `amazing_attribute`. However, I noticed the "old" `amazingobject` in the GIF. I deleted it and inserted a short summary of the result.

PR #370, opened 2023-10-11
Fixes #367 - Edited core worklows

### LINKED ISSUE #367: Edit wording in core-workflows > how-do-they-work
Rework text, examples:
------------------------------------------------

Zammad decides in between selected and saved conditions. These can be combined wherever needed.

    Tip

    🤓 Combining conditions allows “OR”-selections

    However, note that each condition type counts as and selector and can’t overrule the other condition type.

    Every attribute can be used once per condition type.

-----------------------

🧐 Actions can cause confusion
Please note that actions may or may not restrict API based access to attributes. We’re displaying the following icons for your overview to understand these limits better. 👀

ralf401:


system/core-workflows/how-do-they-work.rst
```diff
@@ -189,15 +190,20 @@ unset readonly |ui|
...
-Stop evaluation of other, following workflows that would match otherwise.
+Here you can decide if other workflows are executed after the current one.
+
+If set to ``no`` (default), no further worflows will be executed after the
+current one.
 
-Default: ``no``
+If set to ``yes``, further workflows will be executed if they match the
+condition. In this case, it is possible that your actions from the current
+workflow can be overwritten by another workflow.
```
rolfschmidt, line 200: this looks odd. I think it is the other way around :D

yes will stop
no will continue 


```
If set to ``no`` (default), further workflows will be executed if they match the
condition. In this case, it is possible that your actions from the current
workflow can be overwritten by another workflow.

If set to ``yes``, no further worflows will be executed after the
current one.
```

system/core-workflows/how-do-they-work.rst
```diff
@@ -20,72 +19,71 @@ This will decide on your available conditions and actions.
...
-      evaluated for **all roles**. This also affects your customers! 🙀
+      By default and unless configured in conditions, workflow rules are
+      executed for **all roles**. This also affects your customers!
 
 Selected Conditions
-   These conditions only match if they're active in selection.
-   This applies for drafts (active selection) and currently saved values.
+   These conditions only match if a complying selection is made (e.g. choosing
+   another group in the ticket without saving).
+   This applies also for the the creation of items and temporarily saved values.
```
rolfschmidt, line 56: the old wording was a bit more correct.

Selected conditions are based on the current form values (unsaved changes) on top of the currently saved values of the ticket as basement if the ticket is already stored.

rolfschmidt (changes requested):
see comments

system/core-workflows/how-do-they-work.rst
ralf401: I changed the wording of the block. If it is still not good, please let me know.

rolfschmidt (approved):

PR #381, opened 2023-10-25
Maintenance: Added External Data Source with example and replaced screenshot

ralf401:


ralf401:
@dvuckovic can you have a final look or should I merge directly?

dvuckovic (approved):

PR #396, opened 2023-11-06
Fixes #394 - extended scheduler and data privacy section by objects & new deletion tasks

### LINKED ISSUE #394: Add new data-retention rules for the cleanup of users and associated tickets
```[tasklist]
### Manage > Scheduler
- [x] Reword all paragraphs so they talk about objects generically instead of tickets in particular
- [x] Document the new object attribute where the admin can choose which objects are affected by the job: can now be `User`, `Organization` or `Ticket`
- [x] Update the note about `Action: Delete` for ticket with new name (`Delete immediately`) and additional action called `Add a data privacy deletion task`
- [x] Add an explanation in a box about the difference between `Delete immediately` and `Add a data privacy deletion task` actions, possibly under `Execute changes on objects`
- [x] Update screenshots to show the new object dropdown
```

```[tasklist]
### Misc > Object conditions > Basic object conditions
- [x] `attribtues` => `attributes`
- [x] Add special field: `Existing members (Scope: Organization)`, `Does matching organization has members (or not)?`, `Available matching: is or is not`, `Offered values and options: yes, no`
- [x] Add special field: `Existing tickets (Scope: Ticket Customer, Ticket Owner)`, `Does the matching user has any (open) tickets assigned to them (or not)?`, `Available matching: is or is not`, `Offered values and options: yes, no`
```

```[tasklist]
### System > Data Privacy
- [x] Add a note in the introduction and explain that it is possible to delete matching users and tickets via data privacy tasks from within Scheduler jobs too and link to that section
- [x] `Deleting Users` => `Deleting Users via GUI`
- [x] Add a new Frequently Asked Question that asks about deleting tickets which were re-assigned and explain only the tickets assigned to the matching user at the time of the execution of the data privacy deletion task will be deleted, not on historical basis
- [x] Add a new Frequently Asked Question about extra Delete Ticket tasks shown in the list and explain they are coming from Scheduler jobs when the action was to set to add a data privacy deletion task instead of immediate deletion
```

ralf401:


manage/scheduler.rst
```diff
@@ -3,61 +3,83 @@ Scheduler
...
+   Use :doc:`/system/variables` to build highly-customized email templates.
+
+   Schedulers only perform ``2000`` objects per run. This is a security
+   function in case you accidentally misconfigured the scheduler.
 
 Add a new scheduler
 ===================
 
+Create a new scheduler job by clicking on the "New scheduler" button in the top
+right corner. Then Zammad presents a dialogue where you can create your own one.
```
dvuckovic, line 28: `dialogue` => `dialog`

_Dialogue_ is mostly used for conversations, while _dialog_ is the modal element we all learned to love :)

manage/scheduler.rst
```diff
@@ -3,61 +3,83 @@ Scheduler
...
+
+Example:
+
 .. figure:: /images/manage/scheduler/scheduler-change-owner-in-case-of-ticket-escalation.png
+   :alt: Screenshot showing dialog of creating a new scheduler job
+   :scale: 70 %
```
dvuckovic, line 34: I'm not completely sure, but I think spaces between the value and percentage sign are not allowed.

manage/scheduler.rst
```diff
@@ -3,61 +3,83 @@ Scheduler
...
+   .. warning::
+
+      🔥 Schedulers with *Action: Delete immediately* and *Action: Add a data
+      privacy deletion task* are dangerous and should be used with care! If
+      executed, the objects are deleted and no rollback is possible.
+
+   .. hint::
+
+      **Delete immediately** will delete the object at the runtime of the job
+      *without any hint* in the UI.
```
dvuckovic, line 78: In case of `Delete immediately` we can positively say it relates to ticket objects only. In understand the agnostic approach we agreed on, but this is an exception that proves the rule :)

manage/scheduler.rst
```diff
@@ -3,61 +3,83 @@ Scheduler
...
+
+      **Add a data privacy deletion task** will delete the object at
+      the runtime of the job as a data privacy task, which means it is visible
+      in Zammad's :doc:`data privacy panel </system/data-privacy>`.
```
dvuckovic, line 82: The action does not technically delete the object, it schedules the data privacy deletion task for the aforementioned objects. Then, once the data privacy deletion task is executed, the object is finally deleted. This can take up to 10 minutes on standard installations, which may give some time to the admin to react in case of mistakes.

manage/scheduler.rst
```diff
@@ -66,7 +88,7 @@ Disable notifications
 Note
    You can use the note field to describe the purpose of the scheduler.
    This is only visible to other admins when they are editing the scheduler.
-   It is *not* a way to add notes to tickets.
+   It is *not* a way to add notes to objects.
```
dvuckovic, line 91: Again, the note action is exclusive to the ticket object, so perhaps we can leave this distinction in place.

misc/object-conditions/basics.rst
```diff
@@ -123,29 +144,47 @@ Customer *(Scope: Ticket)*
...
 
-      This field only affects the ticket organization which is set during ticket
-      creation and depends on the ticket customers organization.
+      This field is only available for Scheduler conditions!
 
    .. container:: cfloat-left
 
-      | What ticket organization is affected (or not)?
-      | Available matching: *is*, *is not* or *has changed*
+      | Does the matching organization has members (or not)?
```
dvuckovic, line 154: `has` => `have`

misc/object-conditions/basics.rst
```diff
@@ -123,29 +144,47 @@ Customer *(Scope: Ticket)*
...
+      X
+
+Existing tickets *(Scope: Ticket Customer, Ticket Owner)*
+   .. hint::
+
+      This field is only available for Scheduler conditions!
+
+   .. container:: cfloat-left
+
+      | Does the matching user has any (open) tickets assigned to them (or not)?
```
dvuckovic, line 177: `has` => `have`

system/data-privacy.rst
```diff
@@ -118,26 +125,36 @@ Frequently Asked Questions
...
    When removing users and their tickets, all references are removed.
    This also affects e.g. Reporting - these information are lost.
 
-🤔 How long does Zammad store created tasks?
+How long does Zammad store created tasks?
    Please see the on-premise data section of the
    :docs:`data privacy </appendix/privacy.html>` chapter.
+
+What about re-assigned tickets? I want to delete them, too.
+   Only tickets assigned to the matching user at the time of the execution of
```
dvuckovic, line 153: In wonder if we should be more specific that this only applies to customer assignments, not owners 🤔 Perhaps we could just change to `matching user` to `matching customer` instead.

Remember, deletion tasks will reassigned owner tickets to the system user, they will not be deleted!

dvuckovic (approved):

manage/scheduler.rst
ralf401: At least it works this way, too ;-) But I remove it to have it clean.

manage/scheduler.rst
ralf401: This note field is present for any object selection and should be used for comments on the purpose of the scheduler. So I think it should be fine, or am I wrong?

dvuckovic: Whoops, you're absolutely right, I had Note action in my mind, not Note field. Carry on, nothing to see here!

PR #413, opened 2023-11-24
Fixes #407 - Added custom ticket states

### LINKED ISSUE #407: Add new ticket state management
The new ticket state management area should be described (https://github.com/zammad/zammad/issues/4862):

- Where you will reach it?
- Add/Edit (maybe explain the types?)
- Set default for a new ticket and for follow-up

We should remove the rails console description or do we want to stay with this? (I think in the future this should not be the way to go)

ralf401:


system/objects.rst
```diff
@@ -88,15 +88,97 @@ to Zammad. If you made a mistake or just want to discard your changes, click
...
+        - **merged**: for states for tickets that are merged with other tickets
+        - **new**: for states for tickets that are new and it hasn't been
+          worked on them
+        - **open**: for states for tickets that are in progress and agents are
+          working on them
+        - **pending action**: for states for tickets that are waiting for a
+          specified time and then change their state (default example:
+          *pending close*)
+        - **pending reminder**: for states for tickets that are in progress and
+          you want to set a reminder. (default example: *pending reminder*)
```
mgruner, line 165: I would suggest to order them logically according to ticket flow for easier understanding (new, open, pending, closed, merged). What do you think?

system/objects.rst
```diff
@@ -88,15 +88,97 @@ to Zammad. If you made a mistake or just want to discard your changes, click
 System attributes
 -----------------
 
-Zammad comes with pre-configured attributes. Some of these currently do not
-provide the possibility to edit them via UI (or at all).
+Zammad comes with pre-configured attributes. Some of them can't be edited via
+UI (or at all).
 
 This is not a bug but is to save you from possibly nuking Zammad.
```
mgruner, line 94: How about rephrasing: `This is required for proper operation of Zammad and not a bug.` or similar. I would also connect it to the previous paragraph, it doesn't need an own one.

system/objects.rst
```diff
@@ -88,15 +88,97 @@ to Zammad. If you made a mistake or just want to discard your changes, click
...
 
    .. tip::
 
       There are technical exceptions which can be solved via console.
-      This e.g. affects ticket states and priorities, see
-      :docs:`console section </admin/console.html>`.
+      See :docs:`console section </admin/console.html>` for further information.
 
       💰 If you're a hosted customer, please contact your support for more. 💰
```
mgruner, line 101: Is this still relevant? Perhaps we should clarify this. We actually did the new feature to reduce support requests.

mgruner (approved):
Minor suggestions, please check my notes.

One addition: Would it be nice to have an example use case for adding custom states?

system/objects.rst
ralf401: I also thought about it and kept it aligned to the selection in the UI. But you are right, it is better to have them in logical order.

mgruner: Maybe we should consider changing it also in the UI. Could be a UAT feedback I guess.

mgruner: Alphabetic sorting is not useful at all.

system/objects.rst
ralf401: I removed that part. I think none of the other ticket attributes should (and can?) be manipulated via console.

PR #489, opened 2024-03-11
Add a fix to "CSRF token verification failed!"

AnaelMobilia:
Hello,

When using an apache proxy and using LDAP, "CSRF token verification failed!" occurs. Add to the documentation information from release notes of Zammad 3.2 / https://community.zammad.org/t/zammad-3-2-is-available/3656 .

Bests regards,
Anael


MrGeneration:
The CSRF token error is not bound or specific to LDAP at all. It technically belongs into the system documentation where it already is handled indirectly.

AnaelMobilia:
@MrGeneration  Thanks for outstanding this point!
I will open a PR against this repo.

Bests regards,
Anael

PR #505, opened 2024-03-15
Fixes #502 - Added information about supported group hierarchy levels 

### LINKED ISSUE #502: Group level extension to 10
Only for PSQL, not possible in mySQL!

ralf401:


ralf401:
I just put a sentence to the existing text. I think that's enough, isn't it?

manage/groups/settings.rst
```diff
@@ -22,6 +22,9 @@ Parent group
    permissions. That means you can treat each group as an individual group -
    even if they have a child/parent relation.
 
+   Zammad supports 10 levels of hierarchy for groups on PostgreSQL systems.
+   If you are using a MySQL database, it is limited to 6 levels.
+
```
dvuckovic, line 27: How about adding a hint about migrating to PostgreSQL for existing systems? Like "Did you know that Zammad will deprecate MySQL with version..." with a link to appropriate section on migration. Just a thought, since we should try to communicate in a clear way that we had enough with these MySQL limitations :)

dvuckovic (approved):

PR #527, opened 2024-04-16
Fixes #523 - Rephrase bug note for user search limitation

### LINKED ISSUE #523: Admin > Manage > Users: Search limited to 50 users is not a bug
### Actual behavior
* The documentation states that the limit of search results to 50 users in Admin > Manage > Users, is a bug. 
(See note box at: https://admin-docs.zammad.org/en/latest/manage/users/via-the-admin-panel.html#view-from-users-perspective [scroll up a little])
The note refers to the following issue: https://github.com/zammad/zammad/issues/2214
* In the issue, it is explicitly stated that this hard-coded limit is not regarded as a bug, but bad UI/UX and should be improved.

### Expected behavior
* The note in the admin documentation should reflect how we classify this behavior and if and how we plan to improve it without raising false expectations.

ralf401:


ralf401:
Hi @mgruner I just want to be sure that we can write it that way because I guess it is an old "issue" and there was already some talk about that in the chat...

mgruner (approved):

PR #565, opened 2024-06-06
Add OpenSSL command for Keycloak SAML integration

svenseeberg:
Zammad does not accept the SAML key pair generated by Keycloak, see https://github.com/zammad/zammad/issues/5093. This PR documents an openssl command that generates a key pair that works for Keycloak and Zammad.


ralf401:
Hi @svenseeberg, thanks a lot for your PR! Your solution looks quite nice. However, we discussed it internally and came to the conclusion that we don't want to provide copy & paste commands. We think an admin who has to deal with something like Keycloak should somehow know what he/she is doing. 

Therefore I will add some information about the needed certificate attributes in [this issue](https://github.com/zammad/zammad-admin-documentation/issues/567) and close your PR.

PR #636, opened 2024-08-22
Fixes #630 - Reworked 'Additional follow-up detection' and added new setting

### LINKED ISSUE #630: Document additional follow up feature for email channel
image

ralf401:


channels/email/settings.rst
```diff
@@ -22,14 +22,36 @@ Notification Sender: Default value ``Notification Master <noreply@#{config.fqdn}
...
+
+   Subject & References
+      Checks subject (ticket hook & number) and message IDs of emails. If
+      one of them matches with existing ones, an email is considered as a
+      follow-up.
+
+   References
+      This can be used as an alternative to "Subject & References" if
+      you don't want to have the subject of the email checked but only message
+      IDs.
```
dvuckovic, line 38: I think we should reword this, to mention explicitly that check will only match existing articles in the system with the same references (message ID).

dvuckovic (changes requested):

channels/email/settings.rst
```diff
@@ -22,14 +22,34 @@ Notification Sender: Default value ``Notification Master <noreply@#{config.fqdn}
...
+   If you disable all checks, Zammad only checks if a matching ticket hook &
+   number is present in the subject.
+
+   Subject & References
+      Checks subject and message IDs of emails and searches for initial ticket
+      articles with the same subject and message ID. When both are matching,
+      it is considered as a follow-up.
+
+   References
+      It only checks if a message ID from email matches an existing article.
```
dvuckovic, line 36: It's checked in addition to the default subject ticket hook & number check. Sorry, I thought that was clear. Just reformulate the start of the sentence like those below, and it should be fine.

dvuckovic (approved):

PR #670, opened 2024-10-10
Fixes #669 - Added '~~~' as separator if field has more then one value

### LINKED ISSUE #669: Add separator info for csv import
By zammad 6.4 the CSV import will use `~~~` as separators instead of multiple rows with empty columns.

https://github.com/zammad/zammad/issues/5362

Maybe a small note about the separator `~~~` would be nice here.

https://admin-docs.zammad.org/en/latest/manage/users/via-csv-import.html

ralf401:


ralf401:
Could you have a brief look, @rolfschmidt? You can download the built html files [here](https://github.com/zammad/zammad-admin-documentation/actions/runs/11271694021/artifacts/2039470443).

PR #718, opened 2024-11-14
Fixes #704 - Added description for inactive state for some features

### LINKED ISSUE #704: Add detailed description of what "inactive" means for the different objects
We offer on a lot of objects and object types to toggle between "active" and "inactive". It is not always intuitive what the consequences are, if you set something to inactive. We should add that.

- Users: there are some information, but incomplete. e.g. Users are not able to login if inactive.
- Groups: there are some information, but incomplete. e.g. Customers can see tickets in inactive groups.
- Roles: missing
- Organizations: seems fine
- Overviews: missing
- Text modules: missing
- Macros: seems fine
- Templates: missing
- Checklist templates: missing
- Trigger: missing
- Webhook: seems fine
- Scheduler: vague
- Report profiles: missing
- Time accounting activity types: missing, e.g. can it still be selected by agents? what about time that has already been booked? is it visible in reports and the ui?
- Email accounts: review needed. Is there more?
- Email filters: missing
- Signatures: missing, e.g. what happens to groups that have a inactive signature set
- Object attributes: missing
- Core workflows: missing

Either on a central site or on their according documentation sites.

Edit RSC:

Relevant places to add a description:

```[tasklist]
### Tasks
- [x] Users
- [x] Groups
- [x] Roles
- [x] Templates => Skipping because it simply can't be selected/used anymore
- [x] Time accounting activity types
- [x] Signatures => No need for explanation. It simply won't be included in answers. It even appears a warning in group config dialog if group is inactive.
- [x] Check some places from the suggestion again
```

For the other features it should be clear what it means when something is disabled.

ralf401:


ralf401:
@YetAnotherGerrit would you mind to have a look? Against the title of your issue, I just added a few words for each feature...  I don't know if that meets your expectations, but it is pretty hard to write in detail about such tiny (but not unimportant) functions... At least I tried to check what the effect of the different inactive items is.

Feel free to give me a hint if there is something missing from your point of view.

YetAnotherGerrit (approved):
Thank you. That changes already improve most of the pain points.

PR #738, opened 2025-01-07
Fixes #733 - Add M365 Graph channel guide

### LINKED ISSUE #733: Improve and create new Documentation MS365
```[tasklist]
### Tasks
- [x] Request Account in Entra
- [x] Rename current documentation to` Microsoft 365 IMAP Email`
- [x] Create new documentation for `Microsoft 365 Graph Email` (see [here](https://github.com/zammad/coordination-scrum/issues/221#issuecomment-2567649400) for required steps)
- [x] Create a parent page and link to IMAP and Graph configuration
- [x] Check if duplicate content can be moved to the parent page
- [x] Document 2 ways: shared mailbox and user mailbox
- [x] Off topic: rename Google channel according to Zammad UI
- [x] Separate it according to review / talk with DV
- [x] Replace screenshot of M365 Graph index
- [x] Replace screenshot of M365 IMAP index
- [x] Add error messages
- [x] Replace links in the app - DV
```

ralf401:


channels/microsoft365/m365-graph/index.rst
```diff
@@ -0,0 +1,129 @@
...
+
+- Go to "API permissions" and **Add a permission**.
+- Select "Microsoft Graph" and "Delegated permissions".
+- Add the following permissions:
+
+  - ``offline_access``
+  - ``openid``
+  - ``profile``
+- Add the following permissions for **user mailboxes** only:
+
```
dvuckovic, line 60: Please drop these lines for now. We decided to base the scope on a static list of permissions, so all would be needed regardless of the mailbox type.

channels/microsoft365/m365-graph/index.rst
```diff
@@ -0,0 +1,129 @@
...
+  - ``Mail.ReadWrite``
+  - ``Mail.Send``
+- Add the following permissions for **shared mailboxes** only:
+
```
dvuckovic, line 64: Please drop these lines for now. We decided to base the scope on a static list of permissions, so all would be needed regardless of the mailbox type.

channels/microsoft365/m365-graph/index.rst
```diff
@@ -0,0 +1,129 @@
...
+  - Client Secret: *Value* from client secret
+  - Tenant UUID/Name: *Directory (tenant) ID* (not required for shared mailboxes)
+- Click on **Submit**.
+
+Add Account
+^^^^^^^^^^^
+
+Now you can add your account to Zammad. Do so by clicking the **Add account**
+button in the top right corner. Select the correct mailbox type and click
+**Authenticate**. Enter your credentials and confirm the requested permissions
```
dvuckovic, line 116: Perhaps we can mention here that in case of a shared mailbox, its email address *has* to be provided in a separate input box. Also, it will not be validated at this step (beside some basic email address format validation), since the API cannot be used until the authentication flow is successfully completed.

channels/microsoft365/index.rst
```diff
@@ -1,35 +1,50 @@
-Microsoft 365
-=============
+Microsoft 365 Channels
```
dvuckovic, line 1: Keep in mind that repurposing this page now for *both* type of channels will invalidate the existing link to the docs in the app:

```
https://admin-docs.zammad.org/en/latest/channels/microsoft365/index.html
```

For sure, we can adjust it in the app, but this will only become available with the next general release. Therefore essentially breaking the link for all existing instances in the meantime.

We also cannot introduce a redirect, because the route is just repurposed, it didn't go away.

We have a couple of options:
1. Rename this new overview page to another route, and introduce the redirect for the old path to the new IMAP page
2. Leave the old structure in place, and simply add a separate MS365 directory just for the Graph, keeping the old routes intact.
3. A combination of the two.

Let me know what you decide, but we should sync about it before we merge.

dvuckovic (changes requested):

channels/microsoft365/m365-graph/index.rst
dvuckovic: It simplifies both the code and the setup instructions.

channels/microsoft365/index.rst
ralf401: Thanks for your hint. I'm open for each option but I tend to leave it as it is in the PR because:

- No duplicate content with the current option (the stuff like filters, signatures, settings... )
- M365 email stuff is kept together and the admin can choose which way to go 
- No nearly identical top level sections in the left sidebar (although that would represent the structure in the UI, but I'd like to overcome this approach anyway)

And for the current link: We could change it in the app or even leave it as it is. It should be clear for an admin that there are 2 ways, depending on his M365 setup. And if it is not clear to the admin, it is even more important to see that there are 2 ways to connect. So we could use the same link for M365 Graph Email and the admin always starts with the overview page.

What do you think, are these valid arguments or have I missed something?

ralf401:
Hi @dominikklein & @mantas,

can you please have a look at the changes in the PR? @dvuckovic already had a look and suggested to also ask you for feedback. It would be interesting if you could especially have a look at  the common errors section if there is something missing. 
[
You can download the HTML-Zip here](https://github.com/zammad/zammad-admin-documentation/actions/runs/12687545572/artifacts/2406223982).

channels/microsoft365-graph/accounts.rst
```diff
@@ -0,0 +1,126 @@
...
+  admin interface of Zammad. If it is not correct, change it now. Otherwise the
+  setup of the channel will fail.
+- Go to *Channels > Microsoft 365 Graph Email* and click on
+  **Connect Microsoft 365 App**. Copy the provided callback URL.
+
+Configuration
+----------------
+
+Go to Microsoft's `Entra admin center <https://entra.microsoft.com/#home>`_ and
+log in with as an administrator.
```
dominikklein, line 20: "log in with as" sounds wrong? And also it needs not to be an administrator account only, you need only application administrator permission, but maybe only saying "administrator" is easier to understand.

channels/microsoft365-graph/accounts.rst
```diff
@@ -0,0 +1,126 @@
...
+Enterprise Applications
+^^^^^^^^^^^^^^^^^^^^^^^
+
+- Go to *Applications > Enterprise applications* and select your just created
+  app.
+- Go to "Users and groups" and **Add user/group**.
+- Click "None selected" to add a user.
+- Search and select your user and click on **Select** and then **Assign** to
+  save it.
+
```
dominikklein, line 75: Why is this needed, I did not saw the need in my tests.

dominikklein (changes requested):
In general nice work :-) Only some small comments/questions.

Personally I would love some screenshots for some important things, at least to give some better feeling as an addition to the only text description (e.g. finding correct places in entry, but maybe the people are normaly experts in Microsoft world?).
I saw that nothing related to admin consent is mentioned and also nothing about re-authentication.

channels/microsoft365-graph/accounts.rst
```diff
@@ -0,0 +1,126 @@
...
+Follow the steps below and skip not applicable parts.
+
+First Steps
+-----------
+
+- Check your FQDN under *Settings > System > Fully Qualified Domain Name* in the
+  admin interface of Zammad. If it is not correct, change it now. Otherwise the
+  setup of the channel will fail.
+- Go to *Channels > Microsoft 365 Graph Email* and click on
+  **Connect Microsoft 365 App**. Copy the provided callback URL.
```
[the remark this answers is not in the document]
dvuckovic: Only if you already have the app configured. Initially, the button is displayed on its own with a hint.

channels/microsoft365-graph/accounts.rst
dvuckovic: I believe that in the **App registrations** menu, the `**Users and groups** menu item is simply missing. Could be one of those Microsoft quirks 🤷

dvuckovic: It might be possible to also assign users via **App roles** within the **App registrations** menu, but I didn't test it. Still don't grasp the full concept behind these roles, to be honest.

ralf401: At least I could add the channel without configuring anything in the enterprise application section. Honestly, I am not sure what to do now... Drop it or add an `(optional)` to the title?

dvuckovic: Wait, you were able to configure users for an app without going to that section?! How exactly?

ralf401: Summary after writing with DV: it seems that it is not needed (at least not in my simple tests) so I'll drop it.

ralf401:
> I saw that nothing related to admin consent is mentioned and also nothing about re-authentication.

I added 2 small sections at the bottom of the page about it.

> Personally I would love some screenshots for some important things ...

I added now some screenshots. Let me know if you consider them the right ones ;-)

[Download here.](https://github.com/zammad/zammad-admin-documentation/actions/runs/12691224278/artifacts/2407339680)

channels/microsoft365-graph/accounts.rst
dominikklein: Yes I can agree, it's not needed.

dvuckovic (approved):

PR #796, opened 2025-04-04
Fixes #793 - Add new summary feature and setup

### LINKED ISSUE #793: Add guide for activating summary feature

ralf401:


dvuckovic (approved):
@ralf401 Looks good! Probably better idea to split Providers and Features as separate pages, so we can mimic the admin menu items. But should be fine for the initial state.

Question: what about the user side of this? Do we need some hints about the new personal settings screens + summary sidebar?

ralf401:
Thanks for your review! 

> Probably better idea to split Providers and Features as separate pages, so we can mimic the admin menu items. But should be fine for the initial state.

I created such a structure when I started with it but saw that the pages are pretty empty for the current state. But I totally agree, makes sense to split it when the features are implemented.

> Question: what about the user side of this? Do we need some hints about the new personal settings screens + summary sidebar?

You mean for the user docs? There is another task on which I'll start working. I would keep it rather slim as the other parts in the user docs are as well. 

What do you mean with personal settings screen?

dvuckovic:
> What do you mean with personal settings screen?

Under **Appearance**, when the ticket summary feature is active:

Screenshot 2025-04-08 at 09 42 21

It controls the visibility of the summary bar at the end of the article list (for example, if you decide to hide it permanently).

Screenshot 2025-04-08 at 09 42 41

ralf401:
> Under **Appearance**, when the ticket summary feature is active:

Ah ok I missed that, thanks for the clarification. I mentioned it in the user docs now that it can be re-activated there. I'll add a sentence here as well for the admin, I think it is enough then. Let me know if you have a different opinion, otherwise I'll merge it then.

PR #797, opened 2025-04-07
Note archive mode isn't available when keeping messages

goetzk:
This took me some time to analyse and from looking after the fact I see many other people confused by these two settings. I believe This change gives enough of a hint to eliminate that confusion for most people.

### DIFF
--- a/channels/email/accounts/account-setup.rst
+++ b/channels/email/accounts/account-setup.rst
@@ -187,7 +187,8 @@ Keep messages on server
 
    * ``yes``: Zammad marks imported messages as read. With this option, Zammad
      will only import unread messages. This means Zammad may miss messages if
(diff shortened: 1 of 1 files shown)


ralf401:
Hi @goetzk,

thanks for your contribution. I tested it just now and the archive mode was triggered, even with the setting to keep messages on the server. Maybe you can double check again?

goetzk:
> Hi @goetzk,
> 
> thanks for your contribution. I tested it just now and the archive mode was triggered, even with the setting to keep messages on the server. Maybe you can double check again?

Hi Ralf,
Sorry I missed your reply before.

I don't know when I might have another cause to do an import but if there is I will try to test this again - unfortunately I don't have capacity to re test this from scratch at the moment.

PR #811, opened 2025-04-29
Fixes #808 - Added new integer operators

### LINKED ISSUE #808: 6.6: new operators for integer fields
```
is less than
is less than or equal to
is greater than
is greater than or equal to
```

present basically everywhere so core workflow and also trigger, scheduler etc. for **integer fields**

https://github.com/zammad/zammad/pull/5536

ralf401:


ralf401:
Hey @rolfschmidt I just added them in the object condition section. Seems they were not listed in trigger and scheduler section. Ok that way or do you think we should add more information?

rolfschmidt (approved):

PR #880, opened 2025-07-09
Maintenance: Added robots.txt with '/latest/' as only allowed version to index

ralf401:


ralf401:
Hey @dvuckovic, 

@dominikklein pinged me that some people are asking about how to activate the AI features because they saw search results for the `pre-release` version. 

A test search via Google also showed a result of the version `6.3` before any other version, which is not good (if I remember correctly, I just searched for `zammad trigger management`).

So I tried to add a `robots.txt` as described in the [RTD docs](https://docs.readthedocs.com/platform/stable/reference/robots.html).

What do you think, could this work? I can add it to the other repos then as well and create a reminder to check the results in some days/weeks.

ralf401:
Closing it because this should be discussed first. Currently it is no big problem.

PR #909, opened 2025-08-14
Fixes #905 - Added "Writing Assistant Tool"

### LINKED ISSUE #905: Add write assistant to documentation

ralf401:


ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Default Tools
+-------------
+
+Zammad ships some default tools:
+
+- Improve writing
+- Fix spelling and grammar
+- Expand
+- Simplify
```
dvuckovic, line 18: This has not been clarified yet. But it's almost 💯 we will not have this list in the end.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+
+These are useful basic tools for all agents. This is why they are not
+customizable and deletable. If you enable the feature by turning the
+**Writing Assistant** toggle on, these default tools are available to all
+agents.
```
dvuckovic, line 23: Not exactly true: *all* writing assistant tools are customizable and deletable. Incl. the default ones, they will be included as part of the DB seeds. But they can be modified or removed later by the admin similarly how all default ticket overviews + triggers + other objects are mutable.

I suggest we drop this block, unless you want to mention they are not limited to any groups by default.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Note that these built-in tools can benefit from optimizations which can be
+shipped by an update of Zammad.
```
dvuckovic, line 26: Nope: similar to other DB seeds, we will *not* migrate existing text tools, since they could be modified in unforeseen ways. Because of that, only new instances will benefit from prompt updates for existing tools, unless we decide to add new ones in the future.

I suggest we remove this part.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Add Custom Tools
+----------------
+
+You can add your own custom write tools by providing instructions for the AI
```
dvuckovic, line 31: `write tools` => `writing assistant tools`

Let's try to be consistent with the terminology.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+- **Name**: Add a name for the write tool. This is what your agents see when they
+  access the write assistant tools.
+- **Custom instructions**: Here you have to provide the instructions for the AI.
+  To get proper responses, the instructions should be constructed in a certain
+  way:
+
+   - They should be unambiguous and precise.
+   - They can include one or more examples.
+   - They should not include superfluous or redundant information.
+- **Groups**: Define groups which can access the write tool.
```
dvuckovic, line 53: This is similar to the group limitation for the text modules: we will be filtering available tools based on the current ticket group. In case a text tool has no group limits defined (it's an optional field), it will be available for all groups.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+- **Note**: Add a note about the write tool for other admins. This field is
+  **not** considered as context for the AI.
+- **Active**: Set the write tool to active or inactive.
+
+Note that the **Custom instructions** are a very important part of the
+configuration. They get included in the system prompt and tell the AI what
+to do with a given text.
```
dvuckovic, line 60: Maybe we can mention here that template variable are supported? Similar to text modules, a dropdown suggestion can be triggered by typing `::` in the text box. For sure, any placeholder in the form of `#{...}` will be stored and evaluated, there is no requirement to use the autosuggest feature.

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+
+After saving your writing tool, you should test if the result is as expected. In
+case you are not happy, feel free to adjust the instructions until it fits your
+expectations.
+
+Manage Text Tools
+-----------------
+
+In case you already added at least one text tool, you can see a table with
+all your custom text tools. You can **Clone** or **Delete** them via ︙ action
```
dvuckovic, line 70: Not just *custom* tools: *all* tools will be shown in this table eventually (incl. default ones).

ai/writing-assistant.rst
```diff
@@ -0,0 +1,83 @@
...
+menu. To apply changes to an existing text tool, simply click on the row to
+open the dialog, adjust it and save your changes by clicking the ``Submit``
+button. To set a text tool to **inactive**, set the **Active** attribute to this
+value.
+
+Usage for Agents
+----------------
+
+If enabled, agents have access to the default tools as well as the custom tools
+which are available in the group. They simply have to select the text they want
```
dvuckovic, line 80: `in` => `for`

ai/writing-assistant.rst
ralf401: Yeah, my elaborated text is gone after clicking the `Resolve conversation` button :-(

Hmm... my idea was to reflect the current state of the feature. This also applies to most of your other comments. However, I agree that we should either not create the documentation now at all or not document a "half baked" state. I reworked the complete page, trying to address the mentioned points. Maybe you could have a look again. 

Sorry for missing the target completely...

dvuckovic (approved):

ai/writing-assistant.rst
dvuckovic: Looks great now for the start, thanks for taking care! We will need to touch this very soon, I imagine, but one step at a time.

PR #939, opened 2025-09-25
Fixes #885 - Add alias section for M365 Graph channel

ralf401:


ralf401:
https://github.com/zammad/coordination-scrum/issues/885

PR #945, opened 2025-10-01
Feat: Add devcontainer to README.md

ralf401:


ralf401:
@mgruner & @dvuckovic I added some words about the devcontainer to the README.md. Maybe one of you could have a look if that is ok?

Additionally, I removed the translation diagram because it is only translated into two languages, the others are more or less irrelevant, unfortunately.

PR #974, opened 2025-11-05
Fixes #967 - Optimize wording in triggers

### LINKED ISSUE #967: Enhance trigger content
### Time based execution

3 situations when time based triggers are running: `reminder_reached`, `escalation`, `escalation_warning`

### `Selective` action execution

"When at least one field from conditions was updated or article was added and conditions match."

### Trigger Limitations

Especially second point misleading. Probably refers to variables. Either remove it or mention that you have to create an article when using article attributes in the condition.

### System Notifications

Remove section and check if content should me moved somewhere else

ralf401:


ralf401:
@YetAnotherGerrit Against my announcement in the morning, I did some additional changes to the trigger pages. The trigger section was pretty old and I tried to clean it up a bit. Let me know what you think, thanks!

PR #1142, opened 2026-04-15
Fixes #1124 - Add missing system filters

ralf401:


ralf401:
@dvuckovic

I added now (hopefully) all filters and separated them into two sections (`External Services` & `Internal Email Handling`).

Unfortunately, for most of them I was unable to find and/or understand what they are about. However, I used their small description from the code base and rephrased it a little bit. Even without using AI for it, these descriptions can be considered as highly speculative or even hallucinated. :sweat_smile: 

My idea was that it might be enough to just have such a high level description. Otherwise, they would be on a code-like detail level. Looking forward to read your opinion.

Thanks!

channels/email/filters/system-filters.rst
```diff
@@ -1,36 +1,215 @@
...
+
+0000_postmaster_filter_trusted
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This filter removes ``X-Zammad`` headers from untrustworthy sources to prevent
+manipulation of the email processing. Check the trusted channel section
+in the :doc:`header based actions </channels/email/email-headers>`
+documentation for more information.
+
+0006_postmaster_filter_auto_response_check
```
dvuckovic, line 86: Perhaps we can switch these into a definition list, and style the names as inline code? Just an idea.

dvuckovic (approved):

channels/email/filters/system-filters.rst
ralf401: I converted them to definition blocks now, but without inline code markup. This looks to "noisy" in my eyes.

ralf401:
@mantas I converted the sections now to definition blocks as Dusan suggested, but without applying inline code style, this looks to "noisy" in my eyes. 

Did also a check with Copilot. Some parts got rephrased after it checked the wording against the code.

channels/email/filters/system-filters.rst
```diff
@@ -1,36 +1,173 @@
...
+
+0014_postmaster_filter_own_notification_loop_detection
+   Checks if the email is a self-created notification email, and ignores it in
+   such cases to prevent email loops.
+
+0015_postmaster_filter_identify_session_user
+   Identifies a session user via ``x-zammad-session-user-id``. If none is
+   provided or valid, it falls back to sender data and can create a user.
+
+6500_postmaster_filter_identify_sender
```
mantas: Maybe mention that it also works based on `postmaster_sender_is_agent_search_for_customer` to deal with tickets that rare forwarded from agent personal inboxes into the system. To make sure tey show up as coming from customers and not one of the agents.

channels/email/filters/system-filters.rst
```diff
@@ -1,36 +1,173 @@
...
+   up again, the filter can automatically close the ticket, depending on your
+   settings of the Monit integration.
+
+Internal Email Handling
+-----------------------
+
+   Zammad's internal email processing also uses system filters, e.g. to
+   check for a follow-up email or choose the correct customer. Some of these
+   filters are configurable via Zammad's admin settings. The following titles
+   include the complete filter name for easier reference.
```
mantas: It may be good to mention that they're executed in alphabetical order by the name. Sorting filters by their names in the description might help to understand how the chain works too. Since later filters heavily depend on earlier ones to provide data for  them.

mantas (approved):
Added some tiny comments

PR #1217, opened 2026-06-29
Fixes #1212 - Add spam protection with honeypot and CAPTCHA

### LINKED ISSUE #1212: Add spam protection for form channel
- Add a note about API access to form endpoint with configured spam protection
  - When a CAPTCHA provider is turned on, additional param(s) may be expected, depending on the flavor
  - The actual CAPTCHA challenge will require injection of the relevant script and rendering of the widget, before the param(s) can be obtained
  - Obviously, if the honeypot protection is enabled, the associated param should be skipped in the final payload.
- If the target website that uses the form script employs [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP), it will not be enough to just turn on the CAPTCHA provider in the form configuration. Additional rule exemption will have to be made to accommodate external resources like script files, blobs, etc. Actual resources used highly depend on the chosen CAPTCHA flavor, so an empirical approach would be needed to identify all of the pre-requisites.

ralf401:


dvuckovic (approved):
Very nice, straightforward and to the point!

PR #1261, opened 2026-08-05
Fixes #1250 - Update AI provider configuration: multiple providers and  per-feature configuration

### LINKED ISSUE #1250: Add custom AI provider configuration

ralf401:


ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
+Managing Providers
+------------------
+
+The AI provider configuration page shows a table of all configured providers.
+Each row has a status dot, the provider's name and type, and badges that mark
+the provider as the default for a specific purpose. The ︙ menu in the
+**Action** column lets you change the provider's defaults or remove it. See
+:ref:`defaults` for what each badge means. You can add one provider for all AI
+features you want to enable, or add multiple to assign different providers to
+individual features.
```
dvuckovic, line 30: We should rewrite this paragraph with new changes in mind:

- There can be only one default AI provider (blue `Default` badge) that is used (for chat completions) by all features. Unless overridden, ofc.
- The `Semantic search` and `Image text recognition` badges are now considered provider "capabilities", and they can be set/cleared for one provider at a time. It's currently not possible to override this choice per AI feature.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
 Token
-   Add your API token which you got from your provider here. Don't confuse it
-   with the tokens a large language model (LLM) is dealing with when processing
-   a request.
+   The API token issued by your provider. Don't confuse it with the tokens
+   a large language model deals with when processing a request.
+
+Model
+   The model used to generate text. The placeholder shows Zammad's built-in
+   default for the chosen provider. Leave the field empty to use that default.
```
dvuckovic: Perhaps we can include the word `base` in front of model, to signal this is normally used as a fallback. For example: `The base model used to generate text.`

(Technically, this is the model used for chat completion, but we decided not to use this terminology anymore, since generative text is a bit more palatable.)

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
+
+Embedding Model
+   The model used to turn text into numerical form for semantic search. The
+   placeholder shows Zammad's built-in default for the chosen provider. Leave
+   the field empty to use it. Not every provider supports semantic search; the
+   field appears only for the ones that do.
```
dvuckovic: Perhaps we should not shy away from `vectorizing text` terminology, considering it's used in some UI text already. For example: `The model used to vectorize text for semantic search.`

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
-   Only for Azure AI as provider. Add the endpoint where your text generation
-   model is available.
+   Azure AI only. The endpoint where the text generation model is available.
 
-Model
-   Enter the exact name of the model here. This model has to be available on
-   your provider's side. Leave it empty (if not mandatory) to use the default
-   which is shown as a placeholder.
+URL (Embeddings)
+   Azure AI only. The endpoint where the embedding model is available.
```
dvuckovic: This is unsupported ATM, and we prolly won't implement it, considering it's applicable only to *legacy* Azure AI deployments. Most new deployments are provided under OpenAI compatible API.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
-   Note: the field is called **URL (OCR)** for Azure AI and expects a URL where
-   the OCR model is available instead of a model itself.
+The colored dot at the start of each row shows whether the provider has been
+used successfully or whether it has issues:
 
-After providing the information, make sure to click the ``Save`` button and
-activate AI features.
+Green
+   The provider has been used and the last request succeeded. Shows the
+   timestamp of the last successful connection in a tooltip on hover.
```
dvuckovic: This is not completely true: it's the last _status_ update timestamp, because subsequent OK responses won't update the timestamp due to performance reasons.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
+Orange
+   The provider is configured but has not been used yet. It turns green
+   after the first successful request.
 
-In case you want to know more about Zammad AI or Ollama, read on here. Otherwise,
-head over to :ref:`feedback-logs` or check out the AI-powered features and how
-to use them.
+Red
+   The last request to this provider failed. Hover the dot for the error
+   message from the provider, with a timestamp of the first occurrence.
```
dvuckovic: Here the timestamp behavior is the opposite: it's always the last failed request timestamp, since we always show the most recent error.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
-  configuration.
-- Hosted in the EU and compliant with GDPR.
-- No configuration in Zammad needed (for SaaS customers) or only an API key
-  required (for self-hosted customers).
-- Your requests with sensitive information aren't used for training.
-- If something isn't working as expected, you just have one place where you
-  get support.
+Use for semantic search / Do not use for semantic search
+   Marks the provider as the default for semantic search, or clears that flag.
+   Only available for providers that support it.
```
dvuckovic: Try to reconcile this without using term `default`, since it could be confusing. It's basically setting for which extra AI capability the provider should be used for, other than chat completions/generative text.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
-   Using Zammad AI requires a "V2" plan. Check your
-   :doc:`subscription settings </system/subscription>` and consider to switch
-   your plan. If a "V2" plan is active, you can buy AI calls for AI processing.
-   After doing so, Zammad AI is automatically configured as AI provider and
-   you can activate and configure AI features as you like. Check your remaining
-   AI calls from time to time or activate the notification to be alerted when
-   few calls remain.
+Use for image text recognition / Do not use for image text recognition
+   Marks the provider as the default for image text recognition, or clears that
+   flag.
```
dvuckovic: Same same.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
 
-Ollama
-^^^^^^
+.. _defaults:
 
-`Ollama <https://ollama.com/>`_ is an AI tool that allows you to set up your
-own AI server on a machine you control. That means your data is not transferred
-to a third party. Be aware that this only makes sense if you have a powerful
-GPU in your system!
+Defaults
```
dvuckovic: Yeah, I would split this whole section, per instructions above. We should separate default provider from those marked to be used for certain capability/ability.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
+   used instead.
+
+When you mark a provider as the default for a purpose, Zammad clears that
+flag from whichever provider held it before, so the new default takes over
+immediately.
+
+When you delete the last provider that holds a default, Zammad promotes the
+oldest remaining provider that can serve that purpose to be the default. For
+semantic search, the promotion picks only a provider whose provider type
+actually supports it.
```
dvuckovic: This is not the case anymore: semantic search and image text recognition abilities should not be promoted anymore, and they can be cleared completely.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
+
+If no provider is flagged as the default for semantic search and a feature
+needs it, Zammad shows a warning on the corresponding feature page. Pick a
+provider that supports semantic search as the default to clear the warning.
```
dvuckovic: This is still correct as of now, but it might be a good idea to mention that OCR is an optional ability, and can be completely removed from providers in order to save on AI-related costs.

ai/provider.rst
```diff
@@ -1,162 +1,257 @@
...
    :align: center
-   :width: 60%
 
-Downloads
+The choice only affects text generation. Semantic search always goes to the
+semantic search default, and image text recognition always goes to the image
+text recognition default (or the default provider if no image text recognition
+default is set). The ``Provider`` button is only visible to admins who also hold
+the ``admin.ai_provider`` permission, since routing touches the providers you
+set up on this page.
```
dvuckovic: Do not use the term `default` for semantic search and image text recognition abilities here. Try to align it with other parts.

dvuckovic (changes requested):

dvuckovic (approved):

PR #1272, opened 2026-08-14
Fixes #1268 - New provider configuration with model selection

### LINKED ISSUE #1268: Add model selection to AI provider configuration

ralf401:


ai/provider.rst
```diff
@@ -63,27 +63,52 @@ Type
...
+   list, the field is a dropdown. The default option shows the provider's
+   recommended model; pick a different one if you prefer. If the provider
+   does not offer a model list, enter its name manually.
+
+   Azure AI and Zammad AI do not show this field. Azure AI identifies the
+   model through its deployment URL, while Zammad AI chooses the model itself.
+
+   .. tip:: Need different models for different features? Add the same provider
+      twice with a different model selected in each, then assign each connection
+      to its feature via :ref:`per-feature-provider-config`.
```
dvuckovic, line 89: Great point!

ai/provider.rst
```diff
@@ -63,27 +63,52 @@ Type
...
+
+   When you select or enter an embedding model for which Zammad cannot
+   determine these values, you must also provide the following:
+
+   Embedding dimensions
+      The length of the vectors the embedding model produces.
+
+   Context window size
+      The number of input tokens the embedding model accepts at once.
+
```
dvuckovic, line 109: We should perhaps mention that pre-filled values for dimensions + context window size should *not* be modified, unless you know what you are doing. Although there is a use-case where you want to switch this from default (i.e. runtime Ollama configuration), it should mostly stay as recommended.

ai/provider.rst
```diff
@@ -63,27 +63,52 @@ Type
...
 
 OCR Model
-   The model used to extract text from images. Leave the field empty to fall
-   back to the **Model** field. The placeholder shows Zammad's built-in
-   default for the chosen provider. Not every provider supports image text
-   recognition; the field appears only for the ones that do.
+   The model used to extract text from images. The dropdown lists the models
+   the provider offers. The default option falls back to the **Model** field.
+   Not every provider supports image text recognition; the field appears
+   only for the ones that do.
```
dvuckovic, line 120: Hm, where are we missing it, exactly? I mean, Zammad AI will not show it, but I assure you it supports image text recognition alright 😂

dvuckovic (changes requested):
@ralf401 We don't mention anywhere the new modal is now a "wizard" with two steps. There will be a check and model fetch inbetween the two, which may fail if you enter credentials/URL wrong in the first step (or there is an issue with the service ATM). This is very similar to the check done on Submit, but now we are doing it early, not allowing you to advance to the second step at all, if the provider cannot be reached. This is applicable both for new/edit modals, so maybe worth a mention?

dvuckovic (approved):

PR #1277, opened 2026-08-17
Follow-up #1269 - Overhaul outdated information

### LINKED ISSUE #1269: Adjust required API permission for Microsoft (third-party)
Verify first and implement:

`settings/security/third-party/microsoft.rst:58-61` lists three required API permissions: `openid`, `User.Read`, and `Contacts.Read`. Commit `47386c6d4a` (2026-08-05) removed `Contacts.Read` from the code; the scope is now `openid User.Read` only (`lib/omni_auth/strategies/microsoft_office365_database.rb:21`). 

~Open question: should this change go to latest docs too?~ => `develop` only

### LINKED ISSUE #1276: Follow-up #1269 - Remove or replace outdated screenshots/GIFs
> Below the text that desribes the needed permissions there us a video, that still contains instructions to add `Contacts.Read`. 

 _Originally posted by @YetAnotherGerrit in [#1269](https://github.com/zammad/zammad-admin-documentation/issues/1269#issuecomment-5301917199)_

ralf401:


settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+In multi-tenant (``/common``) setups, sign-in can present an email address
+from any domain. Microsoft signals whether the sending domain is verified on
+the account via the ``xms_edov`` ID-token claim. You can let Zammad enforce
+this before it auto-links a signed-in email to an existing user by enabling
+the **Require verified email domain** option on this screen.
+
+When enabled, Zammad only auto-links an account when the ID token carries
+both an ``email`` claim and a ``xms_edov`` claim whose value is ``true``. If
+either is missing or the domain check is not verified, the user can still log
+in, but the account must be linked to an existing user manually.
```
fliebe92: The described failure behaviour doesn't match the implementation. When the domain check fails, `Authorization::Provider#fetch_user` (`app/models/authorization/provider.rb:23-40`) falls through to user creation:

- with **No user creation on logon** = `yes` → `AccountError`: *"The user account does not exist. Please contact your administrator."*
- otherwise → `User.create_from_hash!` runs into `ensure_uniq_email` (`app/models/user.rb:918`) → *"Email address '…' is already used for another user."*

So in exactly the scenario this section is about (the Microsoft email matches an existing Zammad user, domain unverified) the **login attempt fails with an error** — the user does not get in. Manual linking is then only possible after logging in by other means, via *Profile > Linked Accounts*.

Suggested rewrite, which also covers the default behaviour (currently unstated — worth having, since it explains why the option exists at all):

> When enabled, Zammad only auto-links an account when the ID token carries both an `email` claim and an `xms_edov` claim whose value is `true`. Otherwise the auto-link is refused and the login fails with an error message — the user must then log in to their existing Zammad account by other means and link the Microsoft account under *Profile > Linked Accounts*.
>
> While the option is disabled (default), a missing `xms_edov` claim is treated as verified and only an explicit `false` blocks the auto-link.

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+Click on ``Submit`` to save and activate your configuration.
+
+Optional: restrict sign-in to verified email domains
+------------------------------------------------------
+
+In multi-tenant (``/common``) setups, sign-in can present an email address
+from any domain. Microsoft signals whether the sending domain is verified on
+the account via the ``xms_edov`` ID-token claim. You can let Zammad enforce
+this before it auto-links a signed-in email to an existing user by enabling
+the **Require verified email domain** option on this screen.
```
fliebe92: Missing precondition: this option only has an effect when **Automatic account link on initial logon** is enabled. `Authorization::Provider#fetch_user` only calls `find_user` (and therefore the `xms_edov` check) inside `if Setting.get('auth_third_party_auto_link_at_inital_login')` — and that setting is off by default, so for most instances enabling **Require verified email domain** changes nothing.

Suggest adding a sentence with a cross-reference to the existing target in `settings/security/third-party.rst`, e.g.:

```rst
This option only takes effect if
:ref:`Automatic account link on initial logon <automatic-account-linking>`
is enabled — without it, Zammad never links a third-party login to an
existing account by email in the first place.
```

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+     - **App ID**
+     - From **Overview** in the Entra ID portal.
+   * - **Client secret** (value)
+     - **App Secret**
+     - From **Certificates & secrets** > **Client secrets**. Use the *value*, not the *secret ID*.
+   * - **Directory (tenant) ID**
+     - **App Tenant ID**
+     - Optional. Required only when the app uses the **Single tenant only** account type.
+
+Click on ``Submit`` to save and activate your configuration.
```
fliebe92: The activation step got lost here. Submitting the **Microsoft 365 App Credentials** sub-form saves the credentials but does not enable the provider — *Authentication via Microsoft* is a separate switch (`SettingsAreaSwitch` on the `auth_microsoft_office365` setting, with the credentials form as its `sub`). The previous wording had both ("Apply your settings by pressing submit and activate *Authentication via Microsoft*").

Without it, admins can fill in everything correctly and still see no Microsoft button on the login screen. `google.rst` phrases it as "Enter your **Client ID** and **Client secret** here, enable the toggle and click on ``Submit``" — suggest mirroring that:

> Click on ``Submit`` to save your configuration, then enable the toggle for **Authentication via Microsoft** to activate it.

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+
+Optional: restrict sign-in to verified email domains
+------------------------------------------------------
```
fliebe92: Nit (heading): the option doesn't restrict sign-in — users can still sign in, it only governs whether Zammad auto-links the address to an existing account. Something like "Optional: Require a Verified Email Domain for Account Linking" describes the scope more accurately.

Two style points while touching it:
- Every other heading on this page (and in `google.rst`) is Title Case; this one is sentence case.
- The underline is 54 characters for a 51-character title — valid RST, but the other headings match their titles exactly.

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+
+In multi-tenant (``/common``) setups, sign-in can present an email address
+from any domain. Microsoft signals whether the sending domain is verified on
```
fliebe92: Nit: "sending domain" reads like email-channel wording in what is a login context. `xms_edov` is "Email Domain Owner Verified", so something like "whether the domain of the email address is verified for the account's tenant" is closer.

Minor precision point for the same paragraph: the address Zammad matches on comes from the Graph `/me` profile (`info['email']`), not from an `email` ID-token claim. Requesting the `email` optional claim is Microsoft-side guidance so that `xms_edov` is emitted at all. The current phrasing matches the in-product help text, so this is low priority — just noting it in case you want to be exact.

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
-  Multitenant)
-- Accounts in any organizational directory (Any Azure AD directory -
-  Multitenant) and personal Microsoft accounts (e.g. Skype, Xbox)
+- Single tenant only - restrict sign-in to users (or guests) in your
+  own tenant.
+- Multiple Entra ID tenants - allow users in any Microsoft Entra tenant
+  to sign in.
+- Any Entra ID Tenant + Personal Microsoft accounts - allow users in
+  any Microsoft Entra tenant and with personal Microsoft accounts
+  (e.g. Skype, Xbox) to sign in.
```
fliebe92: These labels and descriptions match the current Entra drop-down verbatim (Learn quickstart, last updated 2026-06-15) — good.

Two follow-ups:
- `channels/microsoft365/accounts/register-app.rst` still lists the pre-Entra labels for the same dialog ("Accounts in this organizational directory only (Single tenant)" …) and refers to "the Azure portals home". The two pages now contradict each other; worth a follow-up issue if it's out of scope here.
- That page also states explicitly that "Personal Microsoft accounts only" is not supported. Since Microsoft lists it as a fourth option and Zammad only builds `/{tenant}` and `/common` endpoints (`lib/omni_auth/strategies/microsoft_office365_database.rb:14-19`), an equivalent note here would save readers a failed setup.

settings/security/third-party/microsoft.rst
```diff
@@ -20,92 +23,91 @@ Limitations
...
+either is missing or the domain check is not verified, the user can still log
+in, but the account must be linked to an existing user manually.
+
+For this to work, the two claims must first be requested on the app. In the
+`Microsoft Entra admin center <https://entra.microsoft.com/>`_, go to
+*Entra ID > App registrations*, select your app and under **Manage** choose
+**Token configuration**. Select **Add optional claim**, set the token type to
+**ID token**, add the claims ``email`` and ``xms_edov`` and save. Until both
+are configured, enabling this option in Zammad blocks all Microsoft 365
+account auto-linking by email.
```
fliebe92: I suggest the following:

```suggestion
Optional: Require a Verified Email Domain for Account Linking
-------------------------------------------------------------

In multi-tenant (``/common``) setups, anyone signing in can present an email
address from any domain. Microsoft signals whether the domain of that address
is verified for the account's tenant via the ``xms_edov`` ("email domain owner
verified") claim in the ID token. Enable **Require verified email domain** to
make Zammad insist on that signal before it links a Microsoft login to an
existing Zammad user by email address.

.. note::

   This option only has an effect in combination with
   :ref:`Automatic account link on initial logon <automatic-account-linking>`.
   Without it, Zammad never links a third-party login to an existing account
   by email address in the first place.

With the option enabled, Zammad links a login to an existing user only if all
of the following apply:

- the ID token contains an ``xms_edov`` claim with the value ``true``,
- the ID token contains a non-empty ``email`` claim, and
- that ``email`` claim matches the email address of the account that is about
  to be linked (upper and lower case are treated as equal).

If one of these conditions is not met, Zammad refuses to link the accounts and
the sign-in fails with an error message – either that the email address is
already in use by another user, or, if **No user creation on logon** is
enabled, that the user account does not exist. Affected users can still get
access: they log in to their existing Zammad account by other means and link
their Microsoft account themselves under *Profile > Linked Accounts*.

While the option is disabled (default), Zammad does not require the claim:
Microsoft either reports a verified domain via ``xms_edov`` or omits the claim
entirely, so an existing account is linked as soon as the email address
matches. Only enabling the option makes the verified domain a requirement.

.. warning::

   Both claims must be requested on your app registration first. In the
   `Microsoft Entra admin center <https://entra.microsoft.com/>`_, go to
   *Entra ID > App registrations*, select your app and choose
   **Token configuration** under **Manage**. Select **Add optional claim**,
   set the token type to **ID token** and add the claims ``email`` and
   ``xms_edov``. As long as this is missing, enabling the option blocks all
   Microsoft account linking by email address.
```

settings/security/third-party/microsoft.rst
ralf401: 1. Issue created: https://github.com/zammad/zammad-admin-documentation/issues/1278
2. I wouldn't add a explicit note because the first sentence in this section says: `Please note that Zammad only supports these account types (app dependent):`. Let me know if you disagree.

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
-  Multitenant)
-- Accounts in any organizational directory (Any Azure AD directory -
-  Multitenant) and personal Microsoft accounts (e.g. Skype, Xbox)
+- Single tenant only - restrict sign-in to users (or guests) in your
+  own tenant.
+- Multiple Entra ID tenants - allow users in any Microsoft Entra tenant
+  to sign in.
+- Any Entra ID Tenant + Personal Microsoft accounts - allow users in
+  any Microsoft Entra tenant and with personal Microsoft accounts
+  (e.g. Skype, Xbox) to sign in.
```
fliebe92: Three things in this list, so one combined suggestion to avoid conflicting patches:

1. **Hyphen as a clause separator.** `Single tenant only - restrict sign-in to…` — a hyphen doesn't separate a label from its description. The house style for labelled bullets is bold + colon, e.g. `system/integrations/exchange.rst:51-54`.
2. **Capitalization drifts inside the list.** "Any Entra ID **T**enant" vs. "Multiple Entra ID **t**enants" two lines up.
3. **Coordination in the third bullet.** "allow users **in** any Microsoft Entra tenant **and with** personal Microsoft accounts" forces one noun through two mismatched prepositions.

```suggestion
- **Single tenant only**: restrict sign-in to users (or guests) in your
  own tenant.
- **Multiple Entra ID tenants**: allow users in any Microsoft Entra
  tenant to sign in.
- **Any Entra ID tenant + personal Microsoft accounts**: allow users in
  any Microsoft Entra tenant, as well as users with personal Microsoft
  accounts (e.g. Skype, Xbox), to sign in.
```

Deliberately left alone: the labels mix "Entra ID tenant" with "Microsoft Entra tenant" in the descriptions, which I assume is faithful to Microsoft's own supported-account-types table — worth a glance against the current portal wording, but not something I'd change blind. Same for the lowercase `personal` above.

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
-Provide the requested information as follows and register your app.
+Sign in to the `Microsoft Entra admin center <https://entra.microsoft.com/>`_,
+go to **App registrations** and create a new app. Provide the requested
+information to register your app.
 
-Name:
-   Any meaningful name fitting, this name will be displayed to users
-   trying to authenticate with this app.
+Name
+   Any meaningful name fits, as this name is displayed to users trying
```
fliebe92: Minor repetition — "Any meaningful **name** fits, as this **name** is displayed".

```suggestion
   Any meaningful name fits, as it is displayed to users trying
```

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
+   to authenticate with this app.
 
-Supported account types:
-   Choose one of the above mentioned account types (see Limitations).
+Supported account types
+   Choose one of the account types listed above. The correct account type
+   depends on your use case. If you only want to use the authentication
```
fliebe92: Stray article — "use **the** authentication internally".

```suggestion
   depends on your use case. If you only want to use authentication
```

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
+   redirect URI field in Microsoft's app configuration.
 
-Redirect URI (optional):
-   Select web and provide your callback URL.
    The callback URL looks like this:
-   ``https://zammad.domain.tld/auth/microsoft_office365/callback``
+   ``https://zammad.example.com/auth/microsoft_office365/callback``
+
+Sign-in requires only the ``User.Read`` API permission, which new apps
+already have. You can verify this under **API permissions**, within
```
fliebe92: The comma before "within" splits a single prepositional phrase.

```suggestion
already have. You can verify this under **API permissions** within
```

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
+With the option enabled, Zammad links a login to an existing user only if all
+of the following apply:
+
+- the ID token contains an ``xms_edov`` claim with the value ``true``,
+- the ID token contains a non-empty ``email`` claim, and
+- that ``email`` claim matches the email address of the account that is about
+  to be linked (upper and lower case are treated as equal).
+
+If one of these conditions is not met, Zammad refuses to link the accounts and
+the sign-in fails with an error message – either that the email address is
```
fliebe92: This `–` is the only en dash in the whole repository (the handful of dashes elsewhere are em dashes). Since what follows is an enumeration of the two error messages, a colon fits and keeps the punctuation set consistent:

```suggestion
the sign-in fails with an error message: either that the email address is
```

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
+their Microsoft account themselves under *Profile > Linked Accounts*.
+
+While the option is disabled (default), Zammad does not require the claim:
+Microsoft either reports a verified domain via ``xms_edov`` or omits the claim
+entirely, so an existing account is linked as soon as the email address
+matches. Only enabling the option makes the verified domain a requirement.
+
+.. warning::
+
+   Both claims must be requested on your app registration first. In the
```
fliebe92: "requested on your app registration" reads a little oddly, and the steps that follow are about configuring an optional claim rather than requesting one:

```suggestion
   Both claims must be configured on your app registration first. In the
```

settings/security/third-party/microsoft.rst
```diff
@@ -18,94 +21,119 @@ This works for Azure users as well and can be an alternative to
...
+   `Microsoft Entra admin center <https://entra.microsoft.com/>`_, go to
+   *Entra ID > App registrations*, select your app and choose
+   **Token configuration** under **Manage**. Select **Add optional claim**,
+   set the token type to **ID token** and add the claims ``email`` and
+   ``xms_edov``. As long as this is missing, enabling the option blocks all
```
fliebe92: Number disagreement: the paragraph opens with "**Both claims** must be…", so the referent here is plural.

```suggestion
   ``xms_edov``. As long as they are missing, enabling the option blocks all
```

settings/security/third-party/microsoft.rst
```diff
@@ -1,13 +1,16 @@
...
 =========
 
-Zammad's Microsoft connection allows your users with Microsoft accounts to login.
-This works for Azure users as well and can be an alternative to
-:doc:`/system/integrations/ldap/index`.
+Zammad's Microsoft connection allows your users with Microsoft accounts to log
+in. This works for Entra ID users as well and can be an alternative to
+:doc:`/system/integrations/ldap/index`. The configuration in Zammad is located
+under *Settings > Security > Third-party Applications* in the section
+**Authentication via Microsoft**. To access it, the permission
```
fliebe92: Optional idiom polish — "in the section **X**" reads more naturally as "in the **X** section". Not an error, so feel free to leave it.

```suggestion
under *Settings > Security > Third-party Applications* in the
**Authentication via Microsoft** section. To access it, the permission
```

fliebe92 (approved):

PR #1304, opened 2026-09-14
Maintenance: keep POT-Creation-Date empty to avoid Weblate merge conflicts

suelje:
## Problem

The Weblate components for the documentation repositories repeatedly end up locked with *"Could not merge the repository"*, requiring a manual *Reset and re-apply* by an admin. See the community report in https://community.zammad.org/t/-/20631.

## Root cause

Sphinx' gettext builder stamps the build time into the catalog header (`POT-Creation-Date`) on every `make gettext` run. On the Weblate side, msgmerge copies that value into the header of **every** `.po` file when it aligns the translations with the new catalog.

Weblate holds translation commits locally until they are pushed (lazy commits, also triggered by file downloads). When a documentation PR with a regenerated catalog is merged while such commits are pending, Weblate's next repository update rebases them onto a branch where the very same header line has changed — a content conflict in every language file at once. Weblate's own gettext merge driver could resolve exactly this conflict, but it is intentionally skipped during rebases (`WEBLATE_MERGE_SKIP=1`), so the update fails hard and the component locks itself.

Notably, Sphinx itself already treats this header line as noise: its `should_write` check ignores `POT-Creation-Date` when deciding whether the catalog changed, and only rewrites the file on real content changes. The field carries no signal in this workflow — it only creates diff churn.

## Fix

Keep `POT-Creation-Date` empty, following the existing convention of the `zammad/zammad` i18n catalog:

* The `gettext` make target now normalizes the header after the Sphinx run, so every regeneration produces a deterministic catalog. Since the CI *Check Translation Sources* step runs the same target, it automatically enforces the normalization as well — a catalog committed with a timestamp fails the existing check. The documented contributor workflow (`make gettext`) is unchanged.
* The committed catalog is normalized once in this PR.
* The `.po` files are deliberately left untouched: Weblate's msgmerge run propagates the empty value into their headers through its own commits, which is conflict-free by construction. After that, the header line never changes again on either side.

## Considered alternative: SOURCE_DATE_EPOCH

Sphinx has no option to omit the header line (it is hard-wired in its `message.pot_t` template, which is not overridable via `templates_path`). It does support the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) but only for copyright information string.

## Validation

Reproduced end-to-end against a local Weblate 2026.8.1 instance with this repository's content: pending translation commits plus an upstream `POT-Creation-Date` bump reliably produce the reported rebase conflict and component lock; with the normalized catalog the same cycle runs clean. The normalization itself is idempotent (a second run produces no further diff).

## Rollout

* The same change needs to be cherry-picked to the `main` branch (separate catalog with a different timestamp value — resolve by keeping the empty field).
* The sibling documentation repositories need the equivalent two-file patch (adjust the catalog filename in the `perl` line accordingly).
* Existing timestamps in the `.po` headers disappear with the next msgmerge run on the Weblate side; no action needed.


mgruner (approved):
Looks great, thanks for the investigation!

